PR #1950, opened 2025-10-27
assets: review connections map links

retini:


retini:
@gazerro, @zapateo, as discussed some time ago, in this PR I have updated the links between connection blocks in the connections map according to the following criteria:

- Gray links are displayed for all connections that have at least one action, regardless of whether the action is enabled or has its scheduler active.

- When hovering over a connection block, the link turns purple and displays a continuous dashed animation if the connection has at least one action that is enabled. For actions targeting users, the scheduler must also be active for the animation to appear. If no actions meet these conditions, the link remains gray and static on hover.

- Connections without any actions do not display a link — neither by default nor on hover.

I’ve also opened a non–release-blocker issue (#1951) to track the other part of our discussion, related to improving the hover animation.

PR #2049, opened 2025-12-11
cmd: add `MEERGO_METRICS_ENABLED` env variable for toggling metrics

gazerro:


zapateo (approved):
The change in the file `compose.yaml` is correct, and the metrics are correctly exposed on http://127.0.0.1:2022/metrics when running Meergo with Docker Compose.

I could not test `tools/docker-compose/compose-release.yaml` as we need to make a release first, but the change in this file is identical to the one in `compose.yaml`, so it looks correct and there is no reason why it should not work.

If you need I can revise something more, otherwise this PR is approved for me.

PR #2131, opened 2026-02-06
Implement an SQL checker for MCP using agents

zapateo:
## Note -- Gianluca, February 9, 2026

We'll leave this PR open for a moment. The current implementation imports many dependencies and should be changed so that none are imported. The AGENTS.md file has already been modified accordingly, so the code needs to be modified. Also, consider whether there are other issues to discuss about this.


zapateo:
Closing this PR in favour of https://github.com/meergo/meergo/pull/2152.

PR #2183, opened 2026-04-22
cmd: add support for AWS Parameter Store and Secret Manager

gazerro:
```
cmd: add support for AWS Parameter Store and Secret Manager

Add the -config-store flag to select where Krenalis reads its startup
configuration from:

  -config-store=env:
      read KRENALIS_* configuration values from the environment

  -config-store=aws:<region>:<prefix>
      read supported configuration values from AWS Systems Manager
      Parameter Store under <prefix> in <region>

The default is '-config-store=env:'.

Database credentials can also be read from an AWS Secrets Manager secret
named '<prefix>/db', using the standard AWS JSON format for RDS
PostgreSQL credentials.
```

PR #2184, opened 2026-04-23
tools/kms: add region to `aws`

gazerro:
```
tools/kms: add region to `aws`

Require AWS KMS URIs to include the target region:

  aws:<region>:<key-id-or-arn>

Pass the region explicitly to the AWS SDK when building the KMS client,
instead of relying on the default configuration chain.
```


zapateo (approved):
I've reviewed the code visually, run the tests, and everything seems fine.

We'll have further review of this change when we push it back to `main` and then integrate it into the PR https://github.com/krenalis/krenalis/pull/2183, so we can test everything on AWS.

PR #2189, opened 2026-04-24
krenalis: update `github.com/getsentry/sentry-go` to v0.46.0

gazerro:


zapateo (approved):
Basically, it seems to me that the Sentry changelog doesn't introduce anything in version 0.46.0 that could break the behavior.

However, exhaustively testing the change is complex, especially considering that we haven't worked on this part for a while, and it involves the server, the Admin console, the GitHub Workflow, and Sentry itself.

If there is an issue, we'll probably see it as we use Krenalis and Sentry.

PR #2195, opened 2026-04-27
core: return a clear error message when database is not initialized

zapateo:
## Commit message

```
core: return a clear error message when database is not initialized

Currently, when the PostgreSQL database connected to Krenalis has not
been initialized, this error is shown to the user:

	error: ERROR: relation "metadata" does not exist (SQLSTATE 42P01)

which is unclear and doesn't indicate what needs to be done.

With this change, the error shown to the user in this scenario will
instead be:

	error: the PostgreSQL database has not been initialized; start
	Krenalis with the -init-db-if-empty flag to initialize it
```

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -270,7 +270,7 @@ func New(ctx context.Context, conf *Config) (_ *Core, err error) {
 	if err != nil {
 		// Return a clear error if the database has not been initialized.
 		if dbpkg.IsUndefinedTable(err) {
-			return nil, errors.New("Krenalis's internal PostgreSQL database has not been initialized. Start Krenalis with the -init-db-if-empty flag to initialize it")
+			return nil, fmt.Errorf("%s (Krenalis's internal PostgreSQL may not be initialized. Try starting Krenalis with the -init-db-if-empty flag)", err)
 		}
 		return nil, err
 	}
--- a/core/core.go
+++ b/core/core.go
@@ -47,7 +47,6 @@ import (
 
 	"github.com/getsentry/sentry-go"
 	"github.com/google/uuid"
-	"github.com/jackc/pgx/v5/pgconn"
 	"golang.org/x/crypto/bcrypt"
 )
 
@@ -270,8 +269,7 @@ func New(ctx context.Context, conf *Config) (_ *Core, err error) {
 	core.state, err = state.New(ctx, db, kms, connectorsOAuth, sendStats)
 	if err != nil {
 		// Return a clear error if the database has not been initialized.
-		var pgErr *pgconn.PgError
-		if errors.As(err, &pgErr) && pgErr.Code == "42P01" {
+		if dbpkg.IsUndefinedTable(err) {
 			return nil, errors.New("Krenalis's internal PostgreSQL database has not been initialized. Start Krenalis with the -init-db-if-empty flag to initialize it")
 		}
 		return nil, err
--- a/core/core.go
+++ b/core/core.go
@@ -47,6 +47,7 @@ import (
 
 	"github.com/getsentry/sentry-go"
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


core/core.go
```diff
@@ -47,6 +47,7 @@ import (
 
 	"github.com/getsentry/sentry-go"
 	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5/pgconn"
```
gazerro: I wouldn't create a direct dependency between `core` and `github.com/jackc/pgx/v5/pgconn`. Instead, I'd have `core/internal/db` return a specific error.

core/core.go
```diff
@@ -268,6 +269,11 @@ func New(ctx context.Context, conf *Config) (_ *Core, err error) {
 	}
 	core.state, err = state.New(ctx, db, kms, connectorsOAuth, sendStats)
 	if err != nil {
+		// Return a clear error if the database has not been initialized.
+		var pgErr *pgconn.PgError
+		if errors.As(err, &pgErr) && pgErr.Code == "42P01" {
+			return nil, errors.New("the PostgreSQL database has not been initialized; start Krenalis with the -init-db-if-empty flag to initialize it")
```
gazerro: I would make it explicit that this is Krenalis's own database and not a data warehouse. It could be as simple as:

> Krenalis's internal PostgreSQL database has not been initialized. Start Krenalis with the -init-db-if-empty flag to initialize it

gazerro (changes requested):

core/core.go
zapateo: Done in [Change errore message](https://github.com/krenalis/krenalis/pull/2195/commits/f332e18477ba4aaa484772e5717848c8eb9be45f).

core/core.go
zapateo: Done in [Implement and use IsUndefinedTable](https://github.com/krenalis/krenalis/pull/2195/commits/24eb9f20b1ecfde463eee66a91a031931bc50854).

Please note that the implementation of the `IsUndefinedTable` function is different from that of the `IsForeignKeyViolation` function, as the error is not directly a `*pgconn.PgError`, but a more "outer" error (`*pgconn.PrepareError`) that wraps `*pgconn.PgError` inside:

Schermata del 2026-04-27 14-12-03

So if the implementation had been the same as `IsForeignKeyViolation` (a simple type assertion), the code wouldn't have worked.

This makes me wonder: is `IsForeignKeyViolation` correct as it is now, in `main`? Or does it actually always return `false`?

gazerro: > This makes me wonder: is IsForeignKeyViolation correct as it is now, in main? Or does it actually always return false?

I investigate further and add some tests.

core/core.go
```diff
@@ -268,6 +268,10 @@ func New(ctx context.Context, conf *Config) (_ *Core, err error) {
 	}
 	core.state, err = state.New(ctx, db, kms, connectorsOAuth, sendStats)
 	if err != nil {
+		// Return a clear error if the database has not been initialized.
+		if dbpkg.IsUndefinedTable(err) {
+			return nil, errors.New("Krenalis's internal PostgreSQL database has not been initialized. Start Krenalis with the -init-db-if-empty flag to initialize it")
```
gazerro: If a table is missing due to a failed update or because it was accidentally deleted or renamed, this error message would be misleading and could hide the real issue. It could be changed to something like:

```
fmt.Sprintf("%s (Krenalis's internal PostgreSQL may not be initialized. Try starting Krenalis with the -init-db-if-empty flag)", err))
```

zapateo: I changed the error message in [Change error message](https://github.com/krenalis/krenalis/pull/2195/commits/e4299451d56445f734b5acc42fb3b3db8127eea2).

PR #2199, opened 2026-04-29
cmd,core: secure API key storage and improves validation

gazerro:
```
cmd,core: secure API key storage and improves validation

Previously, API keys were stored in plaintext in the database. They are
now replaced with an HMAC digest derived using a per-installation
pepper. The pepper is generated at setup time and stored encrypted with
the KMS master key.

A CRC32 checksum is also appended to each key token. This enables the
client to detect copy-paste errors early and return a clear,
deterministic validation error.
```


core/internal/state/keep.go
```diff
@@ -871,9 +871,9 @@ func (state *State) deleteOrganization(n notification) uuid.UUID {
 		delete(state.workspaces, id)
 	}
 	// Delete all access keys belonging to the organization.
-	for token, key := range state.accessKeyByToken {
+	for token, key := range state.accessKeyByHMAC {
```
zapateo: The name of the variable `token` should be updated.

core/core.go
```diff
@@ -396,6 +396,35 @@ func New(ctx context.Context, conf *Config) (_ *Core, err error) {
 	return core, nil
 }
 
+// UpgradeDBAddAPIKeyPepper adds the KMS-encrypted API key pepper to the
+// database. Existing access keys are deleted.
+func UpgradeDBAddAPIKeyPepper(ctx context.Context, conf *Config) error {
```
zapateo, line 401: It would be helpful to also add a little feedback to the log when performing this upgrade (like we do when initializing the database), so that we have more feedback when we upgrade Krenalis instances.

core/internal/state/load.go
```diff
@@ -232,12 +235,14 @@ func (state *State) load(ctx context.Context, oauthCredentials map[string]*OAuth
...
 	}
 	notificationKey := state.cipher.Key(kmsEncryptedNotificationKey)
+	clear(kmsEncryptedNotificationKey)
 	validNotificationKeyCh := make(chan error)
 	go func() {
 		validNotificationKeyCh <- notificationKey.IsValid(ctx)
 	}()
-
-	clear(kmsEncryptedNotificationKey)
+	// API key pepper.
```
zapateo, line 243: We validate the `notificationKey` in a goroutine:

https://github.com/krenalis/krenalis/blob/2909eccbefb480362c9cd3922e2c90e581008757/core/internal/state/load.go#L239-L242

And we check its length:

https://github.com/krenalis/krenalis/blob/2909eccbefb480362c9cd3922e2c90e581008757/core/internal/state/load.go#L233-L236

Should we also do this for `kmsEncryptedAPIKeyPepper`? Or is this conceptually unnecessary?

core/internal/state/accesskey.go
```diff
@@ -0,0 +1,148 @@
...
+	encodeFixedBase62(out[accessKeyPayloadBase62Size:], crc[:])
+	return string(out)
+}
+
+// parseAccessKey parses token as an access key and stores the 32-byte payload
+// in dst. It returns errInvalidToken if token is malformed or fails its CRC
+// check.
+//
+// parseAccessKey may write to dst even if it returns an error.
+func parseAccessKey(dst []byte, token string) error {
```
zapateo: This function performs slices without checking the length first, thus panicking if the slice fails.

I reproduced the problem in a Go source, to run directly with `go run main.go`:

```go
package main

import (
	"encoding/binary"
	"errors"
	"hash/crc32"
	"log"
)

const (
	accessKeyPayloadBase62Size = 43
	accessKeyPayloadSize       = 32
	invalidBase62Digit         = 0xff
)

var base62Digits = [256]byte{
	'A': 0, 'B': 1, 'C': 2, 'D': 3, 'E': 4, 'F': 5, 'G': 6, 'H': 7, 'I': 8, 'J': 9,
	'K': 10, 'L': 11, 'M': 12, 'N': 13, 'O': 14, 'P': 15, 'Q': 16, 'R': 17,
	'S': 18, 'T': 19, 'U': 20, 'V': 21, 'W': 22, 'X': 23, 'Y': 24, 'Z': 25,
	'a': 26, 'b': 27, 'c': 28, 'd': 29, 'e': 30, 'f': 31, 'g': 32, 'h': 33,
	'i': 34, 'j': 35, 'k': 36, 'l': 37, 'm': 38, 'n': 39, 'o': 40, 'p': 41,
	'q': 42, 'r': 43, 's': 44, 't': 45, 'u': 46, 'v': 47, 'w': 48, 'x': 49,
	'y': 50, 'z': 51, '0': 52, '1': 53, '2': 54, '3': 55, '4': 56, '5': 57,
	'6': 58, '7': 59, '8': 60, '9': 61,
}

var errInvalidToken = errors.New("access key token is invalid")

func decodeFixedBase62(dst []byte, s string) error {
	clear(dst) // initialize the base62 accumulator
	for i := 0; i < len(s); i++ {
		digit := base62Digits[s[i]]
		if digit == invalidBase62Digit {
			return errInvalidToken
		}
		carry := int(digit)
		for j := len(dst) - 1; j >= 0; j-- {
			v := int(dst[j])*62 + carry
			dst[j] = byte(v)
			carry = v >> 8
		}
		if carry != 0 {
			return errInvalidToken
		}
	}
	return nil
}

func parseAccessKey(dst []byte, token string) error {
	err := decodeFixedBase62(dst, token[:accessKeyPayloadBase62Size])
	if err != nil {
		return errInvalidToken
	}
	var crc [4]byte
	err = decodeFixedBase62(crc[:], token[accessKeyPayloadBase62Size:])
	if err != nil {
		return errInvalidToken
	}
	if crc32.ChecksumIEEE(dst) != binary.BigEndian.Uint32(crc[:]) {
		return errInvalidToken
	}
	return nil
}

func main() {
	payload := make([]byte, accessKeyPayloadSize)
	err := parseAccessKey(payload, "aaa")
	if err != nil {
		log.Fatal(err)
	}
}

```

zapateo (changes requested):
In the meantime, I've noted these things.

Once these are resolved, I'll continue with the revision by testing the change as a whole.

core/internal/state/keep.go
gazerro: Fixed in commit [Fix variable name](https://github.com/krenalis/krenalis/pull/2199/changes/d3e2e86d8a3126c50d3c09902e5eb5d6ae1d2f45).

core/core.go
gazerro: Done in commit [Log database upgrade status](https://github.com/krenalis/krenalis/pull/2199/changes/ae16fa94b2f22116b57ccf3381dccb69bda71e7f).

core/internal/state/load.go
gazerro: I've checked the length in commit [Validate API key pepper when loading state](https://github.com/krenalis/krenalis/pull/2199/changes/afaebe2920bed781a014d0c8d0df1815708ba613).

If you are also suggesting that we should validate whether the key is actually valid, as we do for the notification key, note that this validation is not meant to validate the notification key itself. Its purpose is to validate the KMS master key: specifically, to ensure that the master key has not changed and that it can still encrypt the data keys correctly.

Among all the data keys, the notification key is the only one we know will eventually be used; we cannot make the same assumption for the others. So, since I had to choose one key for this check, I chose the notification key.

core/internal/state/accesskey.go
gazerro: `parseAccessKey` is only called by `State.AccessKey`, which guarantees that the lengths are correct. It is a separate function only because this makes it easier to test.

That said, it is indeed clearer if the whole validation is done in `parseAccessKey`. See commit [Return parsed access key payload from parser](https://github.com/krenalis/krenalis/pull/2199/changes/c68a3241599af06f9a9ea53da8e69a2404e0b451).

gazerro:
Note that I also fixed an issue with `validNotificationKeyCh`: in case of an error, the goroutine would have been left blocked indefinitely while trying to write to the channel: [Fix validNotificationKeyCh length](https://github.com/krenalis/krenalis/pull/2199/changes/0503b235f3cdbad795c43b8f21be477c8ba6f292)

It would also be more precise to cancel the context on error, so that the goroutine can be stopped immediately.

core/internal/state/load.go
zapateo: > I've checked the length in commit [Validate API key pepper when loading state](https://github.com/krenalis/krenalis/pull/2199/changes/afaebe2920bed781a014d0c8d0df1815708ba613).

👍🏻 

 > If you are also suggesting that we should validate whether the key is actually valid, as we do for the notification key, note that this validation is not meant to validate the notification key itself. Its purpose is to validate the KMS master key: specifically, to ensure that the master key has not changed and that it can still encrypt the data keys correctly.

> Among all the data keys, the notification key is the only one we know will eventually be used; we cannot make the same assumption for the others. So, since I had to choose one key for this check, I chose the notification key.

Okay, now I understand the situation better. So I don't think it's worth validating this one either, unless another reason comes to light.

zapateo:
> Note that I also fixed an issue with validNotificationKeyCh: in case of an error, the goroutine would have been left blocked indefinitely while trying to write to the channel: [Fix validNotificationKeyCh length](https://github.com/krenalis/krenalis/pull/2199/changes/0503b235f3cdbad795c43b8f21be477c8ba6f292)

Shouldn't this fix go on `main`, since the problem is already there, regardless of this PR?

If it's just for convenience, it's fine for me to do it here; I'm just asking in case I'm not understanding something.

zapateo (approved):
After an initial code review, I ran the full tests and manually verified that the `-upgrade-db-add-api-key-pepper` option updates a Krnealis instance (my local one). Everything seems to work. 👍🏻 

Performing more in-depth tests on the hashing mechanisms (to see if there's anything missing from a security perspective, etc.) is difficult and would require in-depth study.

I'm setting the PR as ready for the merge. See also [the comment I left you](https://github.com/krenalis/krenalis/pull/2199#issuecomment-4369237466); decide what to do.

core/organization.go
```diff
@@ -318,17 +314,24 @@ func (this *Organization) CreateAccessKey(ctx context.Context, name string, work
...
 		Type:         state.AccessKeyType(typ),
-		Token:        generateAccessKeyToken(),
+		HMAC:         hmac,
 	}
 	createdAt := time.Now().UTC().Truncate(time.Second)
 	err = this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
-		_, err := tx.Exec(ctx, "INSERT INTO access_keys (id, organization, workspace, name, type, token, created_at) "+
-			"VALUES ($1, $2, NULLIF($3, 0), $4, $5, $6, $7)", n.ID, n.Organization, n.Workspace, name, n.Type, n.Token, createdAt)
+		_, err := tx.Exec(ctx, "INSERT INTO access_keys (id, organization, workspace, name, type, hmac, hint, created_at) "+
+			"VALUES ($1, $2, NULLIF($3, 0), $4, $5, $6, $7, $8)", n.ID, n.Organization, n.Workspace, name, n.Type, hmac[:], hint, createdAt)
```
zapateo: Does the use of `[:]` in `hmac[:]` have some meaning I don't understand, or is it redundant?

core/organization.go
```diff
@@ -318,17 +314,24 @@ func (this *Organization) CreateAccessKey(ctx context.Context, name string, work
 	if err != nil {
 		return 0, "", err
 	}
+	// Generate a random access key,
+	token, hmac, err := this.core.state.GenerateAccessKey(ctx)
+	if err != nil {
+		return 0, "", err
+	}
+	defer clear(hmac)
+	hint := token[:4] + "..." + token[len(token)-6:]
```
zapateo, line 323: Is there a reason this code takes 4 characters at the beginning of the token and 6 at the end? Why not do 4 and 4? It's fine for me to leave it that way (4 and 6), I just wanted to know if there was a specific reason or if it was simply one of many choices.

zapateo (changes requested):

core/organization.go
gazerro: It's redundant, likely it was previously an array: [Remove unnecessary slicing](https://github.com/krenalis/krenalis/pull/2199/changes/5262069a8cbbbfb453bbed2d70ebb7ea87296496)

core/organization.go
gazerro: No particular reason. Looking at it now, I'd probably take four characters at the end as well, since it's just a stylistic choice.

zapateo (approved):

gazerro:
I fixed the dialog that appears when adding an access key, where the key was getting cut off: [Widen access key copy dialog](https://github.com/krenalis/krenalis/pull/2199/changes/667ed2187305f2ffceced5d97eea85bc7d21bc39).

gazerro:
> > Note that I also fixed an issue with validNotificationKeyCh: in case of an error, the goroutine would have been left blocked indefinitely while trying to write to the channel: [Fix validNotificationKeyCh length](https://github.com/krenalis/krenalis/pull/2199/changes/0503b235f3cdbad795c43b8f21be477c8ba6f292)
> 
> Shouldn't this fix go on `main`, since the problem is already there, regardless of this PR?
> 
> If it's just for convenience, it's fine for me to do it here; I'm just asking in case I'm not understanding something.

You're right, I've fixed it in `main`: https://github.com/krenalis/krenalis/pull/2203

zapateo (approved):

PR #2200, opened 2026-04-30
all: update Snowflake Go module to V2

### LINKED ISSUE #2185: warehouses/snowflake,connectors/snowflake: update driver to V2
See https://pkg.go.dev/github.com/snowflakedb/gosnowflake/v2#section-readme.

Currently, we use the V1:

https://github.com/krenalis/krenalis/blob/7e9339b6d5ae3b0b5f0e31b6d962345baf3c6de0/go.mod#L41

zapateo:
Resolves #2185.

## Commit message

```
all: update Snowflake Go module to V2

See https://pkg.go.dev/github.com/snowflakedb/gosnowflake/v2#section-readme.

Resolves #2185.
```

## TODO

* [x] review
* [x] run automated tests
* [x] run manual tests (workspace creation, IR, etc.)


connectors/snowflake/snowflake.go
```diff
@@ -170,15 +170,16 @@ func connector(s *innerSettings) driver.Connector {
...
-		DisableTelemetry: true,
+		Account:   account,
+		User:      s.Username,
+		Password:  s.Password,
+		Role:      s.Role,
+		Database:  s.Database,
+		Schema:    s.Schema,
+		Warehouse: s.Warehouse,
+		Params: map[string]*string{
+			"CLIENT_TELEMETRY_ENABLED": new("false"),
```
gazerro: `new("false")` allocates on every execution. What do you think about using a package-level variable instead?

connectors/snowflake/merge.go
```diff
@@ -174,7 +174,7 @@ func merge(ctx context.Context, conn *sql.Conn, table connectors.Table, rows [][
 	// Copy the deleted rows into the temporary table.
 	if len(deleted) > 0 {
 		// Put the deleted rows into the temporary table's stage.
-		_, err = conn.ExecContext(gosnowflake.WithFileStream(ctx, deletedCSV), `PUT file://rows.csv @%"`+tempTableName+`"`)
+		_, err = conn.ExecContext(gosnowflake.WithFilePutStream(ctx, deletedCSV), `PUT file://rows.csv @%"`+tempTableName+`"`)
```
gazerro, line 177: How was this change tested if the connector tests are skipped? We should re-enable those tests first.

gazerro (changes requested):

connectors/snowflake/snowflake.go
zapateo: Yes, that's fine. What do you think about something like that? [Declare new("false") as package-level variables](https://github.com/krenalis/krenalis/pull/2200/commits/71936e0a4ee3331e695f127e79b4223e2f394fc9)

connectors/snowflake/merge.go
zapateo: This test wasn't executed; the change is based on the indication in point 5 of the `gosnowflake` package documentation:

https://pkg.go.dev/github.com/snowflakedb/gosnowflake/v2#section-readme

Considering that the skipped tests hadn't been run for a very long time, even before they were skipped (and in fact, they stopped working due to some other changes we need to understand), I don't see any particular need to run them before this specific commit, since they were already broken for some time.

That said, I see two options:

1. Leave the skipped tests as they are, assuming the change in this commit is correct, and fix them later.

2. Leave this PR pending, and when we fix the skipped tests, resume this PR.

I'm leaning toward solution 1, which allows us to close this PR in the meantime so as not to leave too many changes pending. Are you still leaning toward solution 2? In that case, I'm fine with that anyway.

gazerro (approved):

PR #2202, opened 2026-05-04
all: support OIDC in Snowflake and enable tests on GitHub Workflow

zapateo:
## Commit message

```
all: support OIDC in Snowflake and enable tests on GitHub Workflow

The purpose of this commit is to test Snowflake on the GitHub Workflow
that runs Krenalis tests, where Snowflake tests are currently disabled.

To ensure that Krenalis running on GitHub Workflow authenticates to
Snowflake, the best solution is to use OIDC (OpenID Connect), since it
doesn't require a token or password; GitHub itself provides a temporary
token for Snowflake for use in the Workflow.

For example, see a documentation page for the 'snow' command that
discusses this authentication mechanism:
https://docs.snowflake.com/en/developer-guide/snowflake-cli/command-reference/auth-commands/overview

This is also discussed in the GitHub documentation:
https://docs.github.com/en/actions/concepts/security/openid-connect

To support authentication, it was necessary to:

- Modify the GitHub Workflow to run the Snowflake tests

- Update the secrets in the Krenalis repository

- Create a user (of type "service") called "GITHUB" on the Snowflake
  instance, which enables authentication via OIDC from the Workflow in
  the Krenalis repository.

- Modify the Snowflake test framework to support OIDC

- Modify the Snowflake connector and warehouse driver to also support OIDC.

Regarding this last point, I chose to add a "token" setting, but it can only be
set via the API, and not through the Settings UI; therefore, it wouldn't be
displayed in the Admin.

This is because I haven't been able to identify a valid use case in
which a user, through the Admin, might want to use an OIDC token, since
they expire very quickly and are therefore tied to CI contexts. Adding
an additional field, mutually exclusive with the password, would have
created confusion.
```


test/snowflaketester/snowflaketester.go
```diff
@@ -198,3 +239,49 @@ func generateTestSchemaName() (string, error) {
...
+// given audience. requestURL and requestToken are the values of the
+// ACTIONS_ID_TOKEN_REQUEST_URL and ACTIONS_ID_TOKEN_REQUEST_TOKEN env vars set
+// by the runner when the job has 'permissions: id-token: write'.
+func fetchGitHubOIDCToken(requestURL, requestToken, audience string) (string, error) {
+
+	sep := "?"
+	if strings.Contains(requestURL, "?") {
+		sep = "&"
+	}
+	req, err := http.NewRequest(http.MethodGet, requestURL+sep+"audience="+audience, nil)
```
gazerro: `audience` must be escaped.

test/snowflaketester/snowflaketester.go
```diff
@@ -198,3 +239,49 @@ func generateTestSchemaName() (string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("cannot build request: %s", err)
```
gazerro, line 256: Should `ACTIONS_ID_TOKEN_REQUEST_URL` be treated as a secret? If so, we need to ensure it never appears in any error messages.

test/snowflaketester/snowflaketester.go
```diff
@@ -198,3 +239,49 @@ func generateTestSchemaName() (string, error) {
...
+	}
+	req.Header.Set("Authorization", "Bearer "+requestToken)
+	req.Header.Set("Accept", "application/json")
+
+	client := &http.Client{Timeout: 30 * time.Second}
```
gazerro: I would avoid redirects:

```go
CheckRedirect: func(req *http.Request, via []*http.Request) error {
        return errors.New("unexpected redirect")
    },
```

.github/workflows/go-run-test-commit.yml
```diff
@@ -42,8 +44,18 @@ jobs:
...
 
-      - name: go run ./test/commit -no-snowflake-tests
+      - name: go run ./test/commit
+        # The Snowflake OIDC token is fetched on demand by snowflaketester
+        # using ACTIONS_ID_TOKEN_REQUEST_URL and ACTIONS_ID_TOKEN_REQUEST_TOKEN,
+        # which are exposed automatically by the runner because the job has
+        # 'permissions: id-token: write'. A single pre-fetched token would
+        # expire (~15 min lifetime) before the long test/commit run completes.
         env:
+          KRENALIS_SNOWFLAKE_TESTER_ACCOUNT: ${{ secrets.KRENALIS_SNOWFLAKE_TESTER_ACCOUNT }}
```
gazerro, line 54: Is there a reason why we have

KRENALIS\_**TEST**\_GOOGLE\_ANALYTICS\_API\_SECRET

but then

KRENALIS\_SNOWFLAKE\_**TESTER**\_ACCOUNT?

connectors/snowflake/snowflake.go
```diff
@@ -171,18 +172,25 @@ func connector(s *innerSettings) driver.Connector {
...
 		Role:      s.Role,
 		Database:  s.Database,
 		Schema:    s.Schema,
 		Warehouse: s.Warehouse,
 		Params: map[string]*string{
 			"CLIENT_TELEMETRY_ENABLED": falseStrPtr,
 		},
-	})
+	}
+	if s.OIDCToken != "" {
```
gazerro, line 186: The token does not seem comparable to a password. It expires after 15 minutes, after which the connection is no longer usable.

This is also evident from the fact that it could never be implemented in the Admin interface, as it is not realistic to require a token when creating a connection. Even via API, it is only usable if one accepts that the connector will stop functioning after a few minutes.

gazerro (changes requested):

test/snowflaketester/snowflaketester.go
zapateo: Done in [Avoid redirects in function fetchGitHubOIDCToken](https://github.com/krenalis/krenalis/pull/2202/commits/9bbfd27b43606812ab6676576a7244d41f2b1834).

test/snowflaketester/snowflaketester.go
zapateo: Done in [Escape audience](https://github.com/krenalis/krenalis/pull/2202/commits/b9ba09ef137ca331a9d8b07f5d766a2c1c29a45e).

test/snowflaketester/snowflaketester.go
zapateo: As discussed, the only possible scenario where the token is displayed in a log (I can't find any others) is if an Internal Server Error occurs on GitHub such that the line:

https://github.com/krenalis/krenalis/blob/b9ba09ef137ca331a9d8b07f5d766a2c1c29a45e/test/snowflaketester/snowflaketester.go#L278

prints the token in the Workflow log.

Regarding this, I have these considerations:

1. All of Workflow and all of Krenalis' code is on GitHub, so GitHub bugs could compromise us in many different ways (this would be just one of many).
2. GitHub has a mechanism that hides tokens in the logs, and that too should fail.
3. Even if an attacker knew that GitHub had a failure and captured the token, what would they do with it? I don't know of any way they could cause any damage without also compromising Snowflake or the Krenalis code on GitHub.

In light of this, I see the advantage of having that line print a log, which greatly simplifies our work in the event of a problem with the OIDC mechanism on GitHub.

If you agree with what I said, I'd leave the implementation as is.

connectors/snowflake/snowflake.go
zapateo: In certain scenarios, the OIDC token could even last longer (or never expire? I don't know), perhaps allowing authentication from cloud applications to Snowflake.

So I wouldn't rule out the possibility of exposing it in the Admin console in the future.

Moreover, as discussed, I can't think of another mechanism that would allow us to easily test the connector and warehouse by instantiating them through the APIs.

So I would leave everything as it is now, at least for now, until we find an ideal solution (that covers all scenarios) or if we run into a technical limitation of this solution (which forces us to think of something different).

Do you agree?

.github/workflows/go-run-test-commit.yml
zapateo: Yes, logically, it's as if they represent two different things.

Variables starting with `KRENALIS_TEST_` are read by Go tests or by the Admin and have no other purpose than be used in automated tests.

Variables starting with `KRENALIS_SNOWFLAKE_TESTER_` are variables read by the `snowflaketester` package, which is used both by the tests and to launch temporary Snowflake instances (test/snowflaketester/cmd).

So in one case, the prefix suggests _variables for Krenalis tests_, in the other, _variables for the Krenalis snowflake tester_.

That's what I thought. Does this distinction make sense to you?

Anyway, keep in mind that this doesn't concern this PR. If you propose changing it, we can discuss it, but we'll have to make a separate change.

gazerro (approved):

PR #2208, opened 2026-05-06
connectors/snowflake, warehouses/snowflake: fix `Merge` CSV columns

### LINKED ISSUE #2198: Test `Test_Columns` fails
The test `Test_Columns` fails, so it has been skipped.

Investigate on it, then re-enable it.

**EDIT**: also see `Test_Merge_Query`

gazerro:
```
connectors/snowflake, warehouses/snowflake: fix `Merge` CSV columns

Preserve connector column names in CSV headers so COPY INTO can match
them correctly with MATCH_BY_COLUMN_NAME = CASE_SENSITIVE.

Test_Merge_Query failed because Merge wrote CSV headers using uppercase
column names, but Snowflake matches those headers against case-sensitive
column names.

The Snowflake warehouse intentionally normalizes identifiers to
uppercase as the physical representation of Krenalis-managed tables.
This is internal to the warehouse and is consistent with its quoted SQL
identifiers and CSV headers.

The connector is different: it operates on user-owned Snowflake tables,
so column names must preserve the case reported by the table schema.
Reusing the warehouse uppercase CSV header behavior caused COPY INTO to
miss lowercase or mixed-case quoted columns and load NULL values
instead.

This mismatch was introduced by 03ae9e271, which aligned the Snowflake
connector and warehouse merge code but carried over the warehouse
uppercase column-name behavior to the connector.

Fixes #2198
```

### DIFF
--- a/connectors/snowflake/merge.go
+++ b/connectors/snowflake/merge.go
@@ -214,7 +214,7 @@ func serializeRowsToCSV(columns []connectors.Column, rows [][]any, deleted bool)
 		if i > 0 {
 			b.WriteByte(',')
 		}
-		b.WriteString(strings.ToUpper(c.Name))
+		b.WriteString(csvColumnName(c.Name))
 	}
 	b.WriteString(",$PURGE\n")
 	for i, row := range rows {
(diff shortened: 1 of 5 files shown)


zapateo (approved):
Approved.

Just review the commit message, in particular:

> but connector table columns are quoted and case-sensitive
in Snowflake

PR #2215, opened 2026-05-12
connectors,core/internal/collector: support IPv6 collector IPs

gazerro:
```
connectors,core/internal/collector: support IPv6 collector IPs

Currently, Krenalis returns HTTP 400 Bad Request when the requester IP
from RemoteAddr, X-Forwarded-For, or context.ip is an IPv6 address.

Fix this by allowing the collector to accept IPv6 addresses from
RemoteAddr, X-Forwarded-For, and context.ip.

Remote address masking preserves the existing IPv4 behavior for the
IPv4 sentinel values 255.255.255.255, 255.255.255.0, and 255.255.0.0,
while applying the IPv6 equivalents /128, /48, and /32 to IPv6 request
IPs.

IPv4-mapped IPv6 addresses are normalized to IPv4, and scoped IPv6
addresses are rejected.

Also update Mixpanel and PostHog connector tests to cover successful
IPv6 IP forwarding.
```

### DIFF
--- a/core/internal/collector/decoder.go
+++ b/core/internal/collector/decoder.go
@@ -583,9 +583,7 @@ func (d *decoder) decodeEvent(connectionId int, fallbackToRequestIP bool) (event
 			context["ip"] = d.remoteAddr.stronglyAnonymous
 			locationIP = d.remoteAddr.ip
 		default:
-			if addr.Is4In6() {
-				addr = addr.Unmap()
-			}
+			addr = addr.Unmap()
 			if addr.IsMulticast() {
 				return nil, errors.BadRequest("property 'ip' cannot be a multicast IP address")
 			}
--- a/core/internal/collector/decoder.go
+++ b/core/internal/collector/decoder.go
@@ -52,9 +52,9 @@ type decoder struct {
 	receivedAt time.Time
 	remoteAddr struct {
 		ip   netip.Addr
-		ip32 string // es. 192.168.1.42
-		ip24 string // es. 192.168.1.0
-		ip16 string // es. 192.168.0.0
+		ip32 string // e.g. 192.168.1.42 or 2001:db8:face:12::1
+		ip24 string // e.g. 192.168.1.0 (/24) or 2001:db8:face:12:: (/64)
+		ip16 string // e.g. 192.168.0.0 (/16) or 2001:db8:face:: (/48)
 	}
 	sentAt       time.Time
 	writeKey     string
@@ -176,7 +176,7 @@ func (d *decoder) Reset(r *http.Request) error {
 		clientIP = strings.TrimSpace(clientIP)
 		err = d.parseRemoteAddr(clientIP)
 		if err != nil {
-			return errors.BadRequest("address specified in the 'X-Forwarded-For' header is not a valid IPv4 address")
+			return errors.BadRequest("address specified in the 'X-Forwarded-For' header is not a valid IP address")
 		}
 	} else {
 		// The address wasn't provided through the 'X-Forwarded-For' header,
@@ -184,7 +184,7 @@ func (d *decoder) Reset(r *http.Request) error {
 		host, _, _ := net.SplitHostPort(r.RemoteAddr)
 		err = d.parseRemoteAddr(host)
 		if err != nil {
-			return errors.BadRequest("only IP version 4 is supported")
+			return errors.New("unexpected IP address from RemoteAddr")
 		}
 	}
 
@@ -567,11 +567,12 @@ func (d *decoder) decodeEvent(connectionId int, fallbackToRequestIP bool) (event
 	// IP.
 	if ip, ok := context["ip"].(string); ok {
 		addr, err := netip.ParseAddr(ip)
-		if err != nil {
+		if err != nil || addr.Zone() != "" {
 			return nil, errors.BadRequest("property 'ip' is not a valid IP address")
 		}
-		if !addr.Is4() {
-			return nil, errors.BadRequest("property 'ip' is not an IPv4 address")
+		mapped := addr.Is4In6()
+		if mapped {
+			addr = addr.Unmap()
 		}
 		switch addr {
 		case ip0: // 0.0.0.0
@@ -588,6 +589,9 @@ func (d *decoder) decodeEvent(connectionId int, fallbackToRequestIP bool) (event
 		default:
 			if addr.IsMulticast() {
 				return nil, errors.BadRequest("property 'ip' cannot be a multicast IP address")
+			}
+			if mapped {
+				context["ip"] = addr.String()
 			}
 			locationIP = addr
 		}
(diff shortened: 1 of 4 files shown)


zapateo (changes requested):
Please change the masking of IPv6 addresses as discussed.

gazerro:
> Please change the masking of IPv6 addresses as discussed.

Done in commit [Use /48 and /32 masks for IPv6 collector IP anonymization](https://github.com/krenalis/krenalis/pull/2215/changes/2818e5cb86156c2d8d061174d016d26fcfd8c257).

zapateo:
> > Please change the masking of IPv6 addresses as discussed.
> 
> Done in commit [Use /48 and /32 masks for IPv6 collector IP anonymization](https://github.com/krenalis/krenalis/pull/2215/changes/2818e5cb86156c2d8d061174d016d26fcfd8c257).

The commit message should be updated according to this.

zapateo:
In the documentation on krenalis.com, consider updating the documentation also for IPv4 anonymization using the mask notation (e.g., ` /48 prefix,`) as we do for IPv6. This is just a suggestion; I'm not sure if it's better or not. Do what you think is the best.

core/internal/collector/decoder.go
```diff
@@ -52,9 +52,9 @@ type decoder struct {
 	receivedAt time.Time
 	remoteAddr struct {
 		ip   netip.Addr
-		ip32 string // es. 192.168.1.42
-		ip24 string // es. 192.168.1.0
-		ip16 string // es. 192.168.0.0
+		ip32 string // e.g. 192.168.1.42 or 2001:db8:face:12::1
```
zapateo: I find the field names used here a bit confusing when they concern IPv6 addresses.

Is there perhaps a way to name these fields that no longer refers to the number of bits (which is no longer meaningful since we are now handling both IPv4 and IPv6), but instead refers to the various levels of anonymization at a more abstract level?

I encountered a similar problem below, when these variables are introduced:

https://github.com/krenalis/krenalis/blob/2818e5cb86156c2d8d061174d016d26fcfd8c257/core/internal/collector/decoder.go#L1117-L1120

Overall, I still find it quite a mental struggle to understand what operations are being performed, having to remember, in the case of IPv6, that the bit numbers are not the actual ones but different ones.

What do you think?

core/internal/collector/decoder.go
```diff
@@ -567,11 +567,12 @@ func (d *decoder) decodeEvent(connectionId int, fallbackToRequestIP bool) (event
 	// IP.
 	if ip, ok := context["ip"].(string); ok {
 		addr, err := netip.ParseAddr(ip)
-		if err != nil {
+		if err != nil || addr.Zone() != "" {
 			return nil, errors.BadRequest("property 'ip' is not a valid IP address")
 		}
-		if !addr.Is4() {
-			return nil, errors.BadRequest("property 'ip' is not an IPv4 address")
+		mapped := addr.Is4In6()
```
zapateo: I found the term `mapped` for this variable name a bit generic, especially when referenced a few lines below. What if it were simply called `ip4in6`, like the method name?

I don't know, maybe "mapped" is a very common term among those working with IPs and I'm just unfamiliar with it. Let me know what you think.

core/internal/collector/decoder.go
```diff
@@ -567,11 +567,12 @@ func (d *decoder) decodeEvent(connectionId int, fallbackToRequestIP bool) (event
...
+		if mapped {
+			addr = addr.Unmap()
```
zapateo: There's something here that I want to understand, whether it's intended or not.

In the case of IPv4 addresses mapped to 6, the unmapping occurs before the switch to `addr`. This means that if an IPv6 address representing, for example, `255.255.0.0` is provided, it would effectively be interpreted as anonymization. This behavior isn't clearly stated in the documentation.

* Is this scenario possible?
* If so, does it need to be explicitly stated in the documentation, or is it somehow implicit for those familiar with working with IP addresses?

zapateo (changes requested):
Please also take a look at the other two comments I added in the Conversation section.

gazerro:
> > > Please change the masking of IPv6 addresses as discussed.
> > 
> > 
> > Done in commit [Use /48 and /32 masks for IPv6 collector IP anonymization](https://github.com/krenalis/krenalis/pull/2215/changes/2818e5cb86156c2d8d061174d016d26fcfd8c257).
> 
> The commit message should be updated according to this.

Done.

core/internal/collector/decoder.go
gazerro: `mapped` is correct in my opinion, but `ip4in6` is clearer, especially because the scope is longer than 23 lines of code.

Done in commit [Rename mapped to ip4in6](https://github.com/krenalis/krenalis/pull/2215/changes/f7a393af5c2aa9ac75caad4a277f2bd8fa3b04d8).

(note that the commit that follows makes this commit unnecessary)

core/internal/collector/decoder.go
gazerro: Fixed in commit [Treat IPv4-mapped IPv6 addresses like regular IP addresses](https://github.com/krenalis/krenalis/pull/2215/changes/8c0cb71404106851412e09ace2888e6d2e5ef6a2).

core/internal/collector/decoder.go
gazerro: Done in commit [Clarify code](https://github.com/krenalis/krenalis/pull/2215/changes/2a40d68c38c7bfc8c6f50692c61d791407133be7).

gazerro:
> In the documentation on krenalis.com, consider updating the documentation also for IPv4 anonymization using the mask notation (e.g., ` /48 prefix,`) as we do for IPv6. This is just a suggestion; I'm not sure if it's better or not. Do what you think is the best.

Done in commit krenalis/krenalis.com@968435f.

core/internal/collector/decoder_test.go
```diff
@@ -1323,12 +1482,12 @@ func TestParseRemoteAddr(t *testing.T) {
...
 		if err == nil {
 			ra := dec.remoteAddr
-			if ra.ip32 != "192.168.1.42" {
-				t.Fatalf("normalization: expected %q, got %q", "192.168.1.42", ra.ip32)
+			if ra.identifiable != "192.168.1.42" {
+				t.Fatalf("normalization: expected %q, got %q", "192.168.1.42", ra.identifiable)
 			}
-			if ra.ip24 != "192.168.1.0" || ra.ip16 != "192.168.0.0" {
+			if ra.partiallyAnonymous != "192.168.1.0" || ra.stronglyAnonymous != "192.168.0.0" {
 				t.Fatalf("masked normalization: expected ip24=%q ip16=%q, got ip24=%q ip16=%q",
```
zapateo: This fatal message should be updated as well (it still refers to `ip24`, etc..).

core/internal/collector/decoder.go
```diff
@@ -567,32 +567,33 @@ func (d *decoder) decodeEvent(connectionId int, fallbackToRequestIP bool) (event
...
-			context["ip"] = d.remoteAddr.ip24
+			context["ip"] = d.remoteAddr.partiallyAnonymous
 			locationIP = d.remoteAddr.ip
 		case ip16: // 255.255.0.0
-			context["ip"] = d.remoteAddr.ip16
+			context["ip"] = d.remoteAddr.stronglyAnonymous
 			locationIP = d.remoteAddr.ip
 		default:
+			if addr.Is4In6() {
+				addr = addr.Unmap()
```
zapateo: Technically, we could call the `Unmap` method here regardless of the IP type, as it's explicitly documented that:

> ..., it returns the wrapped IPv4 address. Otherwise, it returns the unmodified IP.

We already call the method directly in the `parseRemoteAddr` function:

https://github.com/krenalis/krenalis/blob/2a40d68c38c7bfc8c6f50692c61d791407133be7/core/internal/collector/decoder.go#L1113

You decide what you think is clearer.

zapateo (changes requested):

core/internal/collector/decoder_test.go
gazerro: Fixed in commit [Fix error message in tests](https://github.com/krenalis/krenalis/pull/2215/changes/e9d143a7a290cae182086d4c8e9a38a05f681bb3).

core/internal/collector/decoder.go
gazerro: Yes, and it's the same approach we use in `parseRemoteAddr`. Done in commit [Call Unmap unconditionally for context IPs](https://github.com/krenalis/krenalis/pull/2215/changes/a038fd4f37e7cf240958eb6001ac18d362616a59).

zapateo (approved):

PR #2217, opened 2026-05-12
core/internal/collector: accept non-standard X-Forwarded-For headers

gazerro:
```
core/internal/collector: accept non-standard X-Forwarded-For headers

Previously, non-standard X-Forwarded-For headers were rejected with a
400 Bad Request error. Some proxies, such as AWS Application Load
Balancer, may append the client port and use bracketed IPv6 addresses.

This change accepts an optional port after the host and bracketed IPv6
addresses, so the following formats are now accepted:

203.0.113.10
203.0.113.10:54321
2001:db8::1
[2001:db8::1]
[2001:db8::1]:54321

The parser is intentionally permissive and may accept additional
formats; only the formats listed above are explicitly supported.
```


zapateo:
Please merge `main` into this PR and resolve conflicts so I can proceed with the review.

PR #2218, opened 2026-05-12
connectors/http: define a dedicated HTTP transport

gazerro:
```
connectors/http: define a dedicated HTTP transport

Replace the direct use of http.DefaultTransport with an explicitly
configured transport instance.

The transport currently mirrors the default Go configuration, but having
a dedicated instance allows future customization of timeouts, TLS
settings and connection handling.
```


zapateo (approved):
I was wondering about the need for this PR; I was about to ask you, and then I saw that there are other changes based on this that require a dedicated HTTP transport—so this PR is approved.

PR #2220, opened 2026-05-12
connectors/http: harden HTTP transport configuration

gazerro:
```
connectors/http: harden HTTP transport configuration

Improve HTTP client transport security by:

- adding response timeouts to avoid hanging requests
- disabling keep-alive connections for one-shot uploads
- disabling implicit proxy usage to avoid unintended traffic routing
```


connectors/http/http.go
```diff
@@ -234,16 +234,15 @@ func ishex(c byte) bool {
 }
 
 var dialer = &net.Dialer{
-	Timeout:   30 * time.Second,
-	KeepAlive: 30 * time.Second,
```
zapateo, line 238: If the intention is to disable the keep-alive probes, I suggest explicitly indicating that by setting a negative value for `KeepAlive`, as indicated [in the documentation](https://pkg.go.dev/net#Dialer):

> 	// If negative, keep-alive probes are disabled.

zapateo (changes requested):

connectors/http/http.go
gazerro: Done in commit [Disable keep-alive probes explicitly](https://github.com/krenalis/krenalis/pull/2220/changes/d4d57ea632c380259dd439482d078b29d196bb45).

zapateo (approved):

PR #2221, opened 2026-05-12
connectors/http: remove unnecessary response body draining

gazerro:
```
connectors/http: remove unnecessary response body draining 

Remove the explicit draining of the HTTP response body.

Since keep-alives are disabled and connections are not reused,
discarding the response body with io.Copy(io.Discard, res.Body) is
unnecessary. Closing the body is sufficient.
```


zapateo:
Please update this PR (merging the changes from `main` and resolving the conflict), so I can then proceed with the review.

zapateo (approved):

PR #2223, opened 2026-05-13
connectors/s3: use custom S3 HTTP client tuned for large transfers

gazerro:
```
connectors/s3: use custom S3 HTTP client

Previously, S3 requests used the default AWS SDK HTTP client. This
change introduces a custom client.

The custom transport disables proxy lookup from environment variables.

NewBuildableClient is used instead of a plain http.Client so the AWS SDK
can still apply its service/default-mode HTTP customizations before the
underlying client and transport are built.
```

### DIFF
--- a/connectors/s3/s3.go
+++ b/connectors/s3/s3.go
@@ -14,6 +14,7 @@ import (
 	"context"
 	_ "embed"
 	"io"
+	"net/http"
 	"regexp"
 	"strings"
 	"time"
@@ -22,9 +23,10 @@ import (
 	"github.com/krenalis/krenalis/tools/json"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsHTTP "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager"
-	s3pkg "github.com/aws/aws-sdk-go-v2/service/s3"
+	awsS3 "github.com/aws/aws-sdk-go-v2/service/s3"
 )
 
 //go:embed documentation/source/overview.md
@@ -94,7 +96,7 @@ func (s3 *S3) Reader(ctx context.Context, name string) (io.ReadCloser, time.Time
 		return nil, time.Time{}, err
 	}
 	client := s3.client(&s)
-	res, err := client.GetObject(ctx, &s3pkg.GetObjectInput{
+	res, err := client.GetObject(ctx, &awsS3.GetObjectInput{
 		Bucket: aws.String(s.Bucket),
 		Key:    aws.String(name),
 	})
@@ -205,7 +207,7 @@ func (s3 *S3) Write(ctx context.Context, p io.Reader, name, contentType string) 
 }
 
 // client returns a S3 client.
-func (s3 *S3) client(s *innerSettings) *s3pkg.Client {
+func (s3 *S3) client(s *innerSettings) *awsS3.Client {
 	cfg := aws.Config{
 		Region: s.Region,
 		Credentials: aws.NewCredentialsCache(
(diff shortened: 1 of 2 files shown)


connectors/s3/s3.go
```diff
@@ -215,8 +217,16 @@ func (s3 *S3) client(s *innerSettings) *s3pkg.Client {
 				"",
 			),
 		),
+		HTTPClient: awsHTTP.NewBuildableClient().
+			WithTransportOptions(func(transport *http.Transport) {
+				transport.Proxy = nil
+				transport.MaxConnsPerHost = 2
+				transport.MaxIdleConns = 2
```
zapateo: Since S3 serves requests over HTTP1/1, 2 might not be enough here, since we set `opts.Concurrency = 2` above, which would block other HTTP connections from other operations the SDK might perform.

I don't know if this problem will arise in practice, but one possibility, to be more caute, would be to introduce a constant for concurrency and use its value incremented by 1 for the maximum number of HTTP connections.

connectors/s3/s3.go
```diff
@@ -215,8 +217,16 @@ func (s3 *S3) client(s *innerSettings) *s3pkg.Client {
 				"",
 			),
 		),
+		HTTPClient: awsHTTP.NewBuildableClient().
```
zapateo: Consider having a shared connection pool between the various S3 connection instances, since the hosts to which the S3 connector connects are probably shared between various Krenalis users.

connectors/s3/s3.go
```diff
@@ -215,8 +217,16 @@ func (s3 *S3) client(s *innerSettings) *s3pkg.Client {
...
+			WithTransportOptions(func(transport *http.Transport) {
```
zapateo: No timeout is set here for receiving HTTP headers from the server (such as `ResponseHeaderTimeout`).

If so, is this intentional, perhaps because the `ctx context.Context` provided to `S3.Reader` is sufficient? Or is there something else I'm missing?

zapateo (changes requested):
Note: I haven't run any manual tests on the branch yet. I do those at the end, once the code has stabilized, to avoid having to do them multiple times.

connectors/s3/s3.go
gazerro: This issue is no longer relevant following the introduction of the single client/transport architecture. See the following message.

connectors/s3/s3.go
gazerro: I implemented a single transport in  [Use a single transport](https://github.com/krenalis/krenalis/pull/2223/changes/36b753fd5c0beeb2fc0a6120a145258efd461a95).

Probably the constants that are currently defined in the code should be made configurable. But for now I'd avoid adding new environment variables. In the future, we could make them configurable through the API and therefore through the Admin console.

connectors/s3/s3.go
gazerro: Done in commit [Set transport ResponseHeaderTimeout](https://github.com/krenalis/krenalis/pull/2223/changes/f06d67ad32c038bdaa2e9f8cc1072011e12a2e2a).

zapateo: I wonder if 3 seconds is actually too short.

I'm thinking about scenarios where perhaps Krenalis runs on an EC2 instance in one region, while the S3 bucket is in another region.

I'm not sure what value to set, but what if we set a higher value (something like 10 seconds)?

We wouldn't lose much in terms of usability, but we'd gain the advantage of avoiding sporadic errors.

What do you think?

connectors/s3/s3.go
zapateo: As discussed, note that the current implementation, which uses `sync.OnceValue`, does not work as expected, and therefore needs to be revisited.

zapateo (changes requested):

connectors/s3/s3.go
gazerro: Fixed in commit [Fix HTTP client instantiation](https://github.com/krenalis/krenalis/pull/2223/changes/8f47d5650d8635f0f15442912efbddcb9a8df8d1).

connectors/s3/s3.go
gazerro: Changed in commit [Increase response header and idle connection timeouts](https://github.com/krenalis/krenalis/pull/2223/changes/2b34f00c254b7f6ef5f3be45cb45d1c32baed6d6).

connectors/s3/s3.go
zapateo: Considering that the implementation relies on the client always being the same at the package level, might it make sense to add a small test that verifies that, when taking the HTTP client multiple times (i.e. calling `httpClient()`), it is indeed the same?

Perhaps it also prevents a future change from breaking this behavior, "silently" compromising the connector's functionality.

If you think it's reasonable, I think it's worth adding; otherwise, it's fine.

zapateo (approved):

zapateo (changes requested):
See the conversation above.

connectors/s3/s3.go
gazerro: Done in commit [Test HTTP client reuse](https://github.com/krenalis/krenalis/pull/2223/changes/019caec3722acb11f6426066b8934df5433e8f36).

zapateo (approved):

PR #2226, opened 2026-05-13
connectors,core/internal/connectors: use `io.ReadSeeker` in file `Read`

gazerro:
```
connectors,core/internal/connectors: use `io.ReadSeeker` in file `Read`

Allow file connectors to use a Reader method accepting io.ReadSeeker,
avoiding temporary file serialization for connectors requiring
io.ReadSeeker or io.ReaderAt access.

The new Reader method receives a value implementing both io.ReadSeeker
and io.ReaderAt, simplifying file connector implementations and avoiding
double serialization when a temporary file has already been created for
decompression.

As a consequence, the parquet connector Read method is now simpler,
easier to follow, and avoids an additional temporary file serialization.
```


core/internal/connections/file.go
```diff
@@ -943,10 +1041,14 @@ func (c storageWriteCloser) CloseWithError(err error) error {
 // encountered will be logged.
 func removeTempFile(fi *os.File) error {
 	err := fi.Close()
-	if err := os.Remove(fi.Name()); err != nil {
-		slog.Warn("core/connectors: cannot remove temporary file", "path", fi.Name(), "error", err)
+	err2 := os.Remove(fi.Name())
+	if err2 != nil {
+		slog.Warn("core/connectors: cannot remove temporary file", "path", fi.Name(), "error", err2)
 	}
```
zapateo, line 1047: Some kind of problem occurs and is logged to stderr.

I'm using the File System connector, importing a Parquet zipped file (users.zip).

Every time I click "Preview" in the pipeline creation screen, the users are shown in the Admin, but this is logged in the stderr of Krenalis:

```
WARN core/connectors: cannot remove temporary file path=/tmp/115243683 error="remove /tmp/115243683: no such file or directory"
```

core/internal/connections/filestorage.go
zapateo, line 1: While manually testing the import of a Parquet file from Filesystem, when I ran the pipeline I always got an error like this, shown in the pop-up in the Admin:

_unexpected EOF_ (or something like that)

So there was no way to import users.

If it helps, consider that I switched to `main`, and the import worked there. Switching back to this branch caused the problem again.

When I tried to debug the problem, it didn't recur, and I couldn't reproduce it, even by retracing the same steps. I don't know what I did differently.

Roughly, I did the following:

1. Created a pipeline to import users from Parquet files (not compressed)
2. Clicked "Run now" in the pipelines screen

I'm writing this down because there's a problem somewhere, but I don't know how to address it. I'm writing this down because maybe you can think of something, otherwise we'll debug it when it comes up again; I don't know what else we could do.

zapateo (changes requested):

core/internal/connections/file.go
gazerro: This issue also occurs on `main`. Commit [core/internal/connections: avoid duplicate close warning in log](https://github.com/krenalis/krenalis/commit/63bf5ec10050618e7b9eb82c662d2454f1a9a21d) fixed it on `main`.

core/internal/connections/filestorage.go
gazerro: This issue was fixed on the `main` branch by the commit [core/internal/connections: fix storage export connector settings](https://github.com/krenalis/krenalis/commit/83417acad83af0d7a9f9010cf29981cb953e3d14). That's why the error does not occur on `main`.

Note that, due to another bug, the error may temporarily disappear after saving the pipeline, only to reappear after restarting Krenalis. This will be fixed separately by #2231.

gazerro:
See also the following commits resulting from the discussion we had:

- [Log file reader close errors without failing import](https://github.com/krenalis/krenalis/pull/2226/changes/5d70c1fa2555a3eea1bc2c12f9ce8b92c8cfbc8f)
- [Test readseeker fast path hides closer and name](https://github.com/krenalis/krenalis/pull/2226/changes/161f3d28fcefe73a155eaba8f4ca9c971538572a)

zapateo (approved):

PR #2227, opened 2026-05-14
connectors/sftp: add SFTP host key validation

gazerro:
```
connectors/sftp: add SFTP host key validation

Add an optional SFTP setting for the server public key and use it to
validate the SSH host key during connection setup.

When a host key is configured, restrict accepted SSH host key
algorithms to those compatible with the configured key, including
RSA-SHA2 variants for RSA keys.
```

### DIFF
--- a/connectors/sftp/sftp.go
+++ b/connectors/sftp/sftp.go
@@ -7,6 +7,7 @@
 package sftp
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"errors"
@@ -33,6 +34,8 @@ var sourceOverview string
 
 //go:embed documentation/destination/overview.md
 var destinationOverview string
+
+var errHostKeyMismatch = errors.New("host key mismatch")
 
 func init() {
 	connectors.RegisterFileStorage(connectors.FileStorageSpec{
@@ -62,11 +65,12 @@ type SFTP struct {
 }
 
 type innerSettings struct {
-	Host     string `json:"host"`
-	Port     int    `json:"port"`
-	Username string `json:"username"`
-	Password string `json:"password"`
-	TempPath string `json:"tempPath"`
+	Host          string `json:"host"`
+	Port          int    `json:"port"`
+	Username      string `json:"username"`
+	Password      string `json:"password"`
+	HostPublicKey string `json:"hostPublicKey"`
+	TempPath      string `json:"tempPath"`
 }
 
 // AbsolutePath returns the absolute representation of the given path name.
@@ -147,6 +151,7 @@ func (sf *SFTP) ServeUI(ctx context.Context, event string, settings json.Value, 
 			&connectors.Input{Name: "port", Label: "Port", Placeholder: "22", Type: "number", OnlyIntegerPart: true, MinLength: 1, MaxLength: 5},
 			&connectors.Input{Name: "username", Label: "Username", Placeholder: "username", Type: "text", MinLength: 1, MaxLength: 200},
 			&connectors.Input{Name: "password", Label: "Password", Placeholder: "password", Type: "password", MinLength: 1, MaxLength: 200},
+			&connectors.Input{Name: "hostPublicKey", Label: "Server public key", Placeholder: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAA...", HelpText: "Strongly recommended. Ensures the SFTP server identity matches this OpenSSH public key.", Rows: 3, MaxLength: 5000},
 			&connectors.Input{Name: "tempPath", Label: "Temporary directory path", Placeholder: "/", Type: "text", MinLength: 0, MaxLength: 1000, Role: connectors.Destination},
 		},
 		Settings: settings,
@@ -243,6 +248,16 @@ func (sf *SFTP) saveSettings(ctx context.Context, settings json.Value, role conn
 	if n := utf8.RuneCountInString(s.Password); n < 1 || n > 200 {
 		return connectors.NewInvalidSettingsError("password length must be in range [1,200]")
 	}
+	// Validate HostPublicKey.
+	s.HostPublicKey = strings.TrimSpace(s.HostPublicKey)
+	if s.HostPublicKey != "" {
+		if n := utf8.RuneCountInString(s.HostPublicKey); n > 5000 {
+			return connectors.NewInvalidSettingsError("server public key length must be in range [0,5000]")
+		}
+		if err := validateHostPublicKey(s.HostPublicKey); err != nil {
+			return connectors.NewInvalidSettingsError("server public key must be a valid OpenSSH public key")
+		}
+	}
 	// Validate TempPath.
 	if role == connectors.Destination {
 		if n := utf8.RuneCountInString(s.TempPath); n > 1000 {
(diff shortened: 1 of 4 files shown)


connectors/sftp/sftp.go
zapateo, line 1: I think that the SFTP connector's overview should be updated to mention the SSH key:

Image

In particular, in the "**What does it require?**" section, which only mentions the username and the password.

What do you think?

connectors/sftp/sftp.go
```diff
@@ -304,17 +319,65 @@ func (client *client) close() error {
...
+	}
+	return []string{key.Type()}
+}
+
+// hostKeyValidator returns an SSH HostKeyCallback that verifies the server
+// host key matches the expected key exactly. Connections using a different
+// host key are rejected.
+func hostKeyValidator(key ssh.PublicKey) ssh.HostKeyCallback {
+	return func(hostname string, remote net.Addr, candidate ssh.PublicKey) error {
+		if key == nil {
```
zapateo: Excluding internal bugs, this condition should never occur, or am I wrong?

In that case, why introduce a defensive programming condition here? Is there an advantage?

connectors/sftp/sftp.go
```diff
@@ -243,6 +248,16 @@ func (sf *SFTP) saveSettings(ctx context.Context, settings json.Value, role conn
...
 		return connectors.NewInvalidSettingsError("password length must be in range [1,200]")
 	}
+	// Validate HostPublicKey.
+	s.HostPublicKey = strings.TrimSpace(s.HostPublicKey)
+	if s.HostPublicKey != "" {
+		if n := utf8.RuneCountInString(s.HostPublicKey); n > 5000 {
+			return connectors.NewInvalidSettingsError("server public key length must be in range [0,5000]")
+		}
+		if err := validateHostPublicKey(s.HostPublicKey); err != nil {
+			return connectors.NewInvalidSettingsError("server public key must be a valid OpenSSH public key")
```
zapateo: This error masks the more specific errors returned by `validateHostPublicKey`, which in turn are returned by `parseHostPublicKey`.

One problem I see is if the key contains options or similar, because it would be nice to see the error:

> public key options are not allowed

rather than a generic:

> server public key must be a valid OpenSSH public key

when in fact the key is perfectly valid.

What do you think?

connectors/sftp/sftp.go
```diff
@@ -341,6 +404,9 @@ func openClient(ctx context.Context, s *innerSettings) (*client, error) {
 	}()
 	c, chans, reqs, err := ssh.NewClientConn(conn, addr, sshConfig)
 	if err != nil {
+		if errors.Is(err, errHostKeyMismatch) {
+			return nil, connectors.NewInvalidSettingsError("server public key does not match the server host key")
```
zapateo: Perhaps this error message can be made clearer:

> server public key does not match the server host key

It tells me there was a mismatch, but here in the message it compares:

* the server **public** key
* the server **host** key

Actually, when we talk about _public_ and _host_ keys, is it clear that one is the one configured in Krenalis and the other is the actual one of the server one is connecting to?

Not sure if it a problem with my understanding, because I'm not familiar with SSH terminology, or is there actually a clarity issue that could be improved.

connectors/sftp/sftp.go
```diff
@@ -304,17 +319,65 @@ func (client *client) close() error {
...
+	if key.Type() == ssh.KeyAlgoRSA {
+		return []string{ssh.KeyAlgoRSASHA256, ssh.KeyAlgoRSASHA512, ssh.KeyAlgoRSA}
+	}
+	return []string{key.Type()}
+}
+
+// hostKeyValidator returns an SSH HostKeyCallback that verifies the server
+// host key matches the expected key exactly. Connections using a different
+// host key are rejected.
+func hostKeyValidator(key ssh.PublicKey) ssh.HostKeyCallback {
```
zapateo, line 340: I'd add a test on this function for the "positive" case, which doesn't seem to be covered by the tests.

To be clear, if this statement in `hostKeyValidator`:

```go
if !bytes.Equal(candidate.Marshal(), key.Marshal()) {
    return errHostKeyMismatch
}
```

is replaced with:

```go
if true { // WRONG
    return errHostKeyMismatch
}
```

all tests in `sftp_test.go` still pass. So I'd add the case where `hostKeyValidator` returns a positive result.

zapateo (changes requested):

connectors/sftp/sftp.go
gazerro: A defensive programming condition should only be introduced in these cases:

a) to prevent data corruption or other irreversible consequences
b) to make a bug immediately obvious when, without the check, it would become extremely difficult to diagnose

Case (a) does not apply here. As for case (b), the runtime would already panic when invoking a method on a `nil` interface, and the only ambiguity would be whether `candidate` or `key` is `nil`. If we want to remove even that ambiguity, it would be enough to call the two `Marshal` methods on separate lines.

So yes, I agree with you.

connectors/sftp/sftp.go
gazerro: Great catch. Fixed in commit [Improve SFTP host public key validation errors](https://github.com/krenalis/krenalis/pull/2227/changes/048ace51368dd7b1fd014baeb1188b2ed51ae671).

connectors/sftp/sftp.go
gazerro: What do you think about the error message "remote server does not match the configured server public key"?

connectors/sftp/sftp.go
gazerro: Done in commit [Add positive SFTP host key validator test](https://github.com/krenalis/krenalis/pull/2227/changes/0da90f2e528fbed4596cc416689b68a7cea05656).

connectors/sftp/sftp.go
gazerro: We could add:

* Optionally, provide the server host key so Krenalis can verify that it is connecting to the expected server.

zapateo: Yes, fine.

connectors/sftp/sftp.go
zapateo: Yes, we can split across multiple lines. Decide for yourself whether it's worth it; I'm fine with that either way.

connectors/sftp/sftp.go
zapateo: Okay, very clear to me.

zapateo (changes requested):

connectors/sftp/sftp.go
gazerro: Done in commit [Update connector documentation](https://github.com/krenalis/krenalis/pull/2227/changes/3354e4109b0ca46a1d51951ffcb93f9c22731f74).

connectors/sftp/sftp.go
gazerro: Done in commit [Split key unmarshalling across multiple lines](https://github.com/krenalis/krenalis/pull/2227/changes/5bd8cd8668c7154c7228bff8bee807e9e32dc3c8).

connectors/sftp/sftp.go
gazerro: Done in commit [Clarify error message](https://github.com/krenalis/krenalis/pull/2227/changes/f9b6210724b5c9d2219275c15c2392b9acd5c859).

connectors/sftp/sftp.go
zapateo: The `key == nil` check should be removed.

zapateo (changes requested):

connectors/sftp/sftp.go
gazerro: Done in commit [Remove key nil validation in hostKeyValidator](https://github.com/krenalis/krenalis/pull/2227/changes/c983df5fedd885dd9255c5b79c7dd3a83c6051c3).

zapateo (approved):

PR #2231, opened 2026-05-15
core/internal/state: avoid cursor resets for formats without settings

gazerro:
```
core/internal/state: avoid cursor resets for formats without settings

Updating a file pipeline whose format has no settings, such as parquet,
causes the pipeline cursor to be reset. This happens because state
incorrectly stores 'Pipeline.FormatSettings' as 'json.Value("null")'
instead of nil after applying pipeline create/update notifications (for
formats without settings, FormatSettings should remain nil.)

Since the format settings then appear to have changed, shouldReload
resets the pipeline cursor.

This commit normalizes JSON null format settings back to nil when
applying pipeline create and update notifications so that unchanged
pipelines preserve their cursor.
```


test/parquet_import_test.go
```diff
@@ -87,9 +86,14 @@ func TestParquetImport(t *testing.T) {
...
-		UserIDColumn: "parquet_id",
-		Format:       "parquet",
-	})
+		UserIDColumn:    "parquet_id",
+		UpdatedAtColumn: "updated_at",
+		Format:          "parquet",
+		Incremental:     true,
+	}
+
+	// Create a File System source connection, with a pipeline that imports from the Parquet file.
```
zapateo: This comment was moved (I think) by mistake, or not?

zapateo (changes requested):

test/parquet_import_test.go
gazerro: Fixed in commit [Fix comments](https://github.com/krenalis/krenalis/pull/2231/changes/982f874625af3bfd1f896e0ff816eb225a660916).

zapateo (changes requested):
The test in this PR conflicts with `main`. Please resolve the conflicts first, then I'll review the merge and approve it.

gazerro:
> The test in this PR conflicts with `main`. Please resolve the conflicts first, then I'll review the merge and approve it.

As dicussed, I reverted the test change in commit [Revert test change](https://github.com/krenalis/krenalis/pull/2231/changes/929d7e13f01e93e752bc09e7c3c588ee58eadcea).

zapateo (approved):

PR #2237, opened 2026-05-19
core/internal/state: improve state load error messages

gazerro:
```
core/internal/state: improve state load error messages

Improve the error messages shown when an error occurs while loading
state. For example:

Before:
  parsing time "2005-12-06T00:00:00Z": extra text: "T00:00:00Z"

After:
  cannot load pipelines: loading filter for pipeline 3711677:
  parsing time "2005-12-06T00:00:00Z": extra text: "T00:00:00Z"
```

### DIFF
--- a/core/internal/state/load.go
+++ b/core/internal/state/load.go
@@ -257,20 +257,14 @@ func (state *State) load(ctx context.Context, oauthCredentials map[string]*OAuth
 	// Read all organizations.
 	state.organizations = map[uuid.UUID]*Organization{}
 	err = tx.QueryScan(ctx, "SELECT id, name FROM organizations", func(rows *db.Rows) error {
-		var id uuid.UUID
-		var name string
 		for rows.Next() {
-			if err := rows.Scan(&id, &name); err != nil {
-				return fmt.Errorf("loading organization %s: %s", id, err)
+			org := &Organization{mu: new(sync.Mutex)}
+			if err := rows.Scan(&org.ID, &org.Name); err != nil {
+				return fmt.Errorf("loading organization %s: %s", org.ID, err)
 			}
-			organization := &Organization{
-				mu:   new(sync.Mutex),
-				ID:   id,
-				Name: name,
-			}
-			organization.workspaces = map[int]*Workspace{}
-			organization.members = map[int]struct{}{}
-			state.organizations[id] = organization
+			org.workspaces = map[int]*Workspace{}
+			org.members = map[int]struct{}{}
+			state.organizations[org.ID] = org
 		}
 		return nil
 	})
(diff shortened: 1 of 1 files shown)


zapateo (changes requested):
The change itself is correct and the error message is clearer now, but I need to understand better this change.

After testing similar cases, I noticed that the problem is broader and appears to involve other cases as well.

For example, if the transformation mapping is invalid, this error is returned:

```
error: json: unable to unmarshal JSON number into Go map[string]string
```

Which makes it very hard to figure out what the problem is.

Moreover, if the input schema is corrupted, Krenalis prints:

```
error: invalid type syntax
```

Which is unclear as well.

Why prioritize the case where the pipeline filter is corrupted? Is there a specific reason?

gazerro:
As discussed, I've added contextual information to the other error messages as well. I intentionally left out certain SQL errors, such as failures occurring while establishing the initial connection.

core/internal/state/load.go
```diff
@@ -212,27 +212,27 @@ func (state *State) load(ctx context.Context, oauthCredentials map[string]*OAuth
 		&kmsEncryptedNotificationKey, &kmsEncryptedAPIKeyPepper)
 	if err != nil {
 		if err == sql.ErrNoRows {
-			return errors.New("row is missing from table 'metadata'")
+			return errors.New("cannot load metadata: no rows found in table; expected exactly one row")
 		}
-		return err
+		return errors.New("cannot load metadata: %s")
```
zapateo: Here I think you wanted to use `fmt.Errorf` and pass the `err` argument.

core/internal/state/load.go
```diff
@@ -285,7 +285,7 @@ func (state *State) load(ctx context.Context, oauthCredentials map[string]*OAuth
 		var org *Organization
 		for rows.Next() {
 			if err := rows.Scan(&id, &organization); err != nil {
-				return err
+				return fmt.Errorf("loading member %d: %s", id, err)
```
zapateo, line 282: Consider this scenario:

1. The first loop successfully reads the first member.
2. The second loop fails to read the member, so the value of `id` is unchanged and still refers to the previous member, which was read successfully.

The error message, however, would refer to the first member, saying there was an error reading it, when in fact the error is in the second.

Or am I mistaken?

zapateo (changes requested):

core/internal/state/load.go
gazerro: Fixed in commit [Use fmt.Errorf instead of errors.New](https://github.com/krenalis/krenalis/pull/2237/changes/eaf72de9ebe785f6aeea3cdb072b61641fb2447c).

core/internal/state/load.go
gazerro: Good catch. As discussed, I moved the declarations of the variables assigned during the scan inside the loops. In one case, I also removed some temporary variables and assigned the values directly to the struct fields: [Move variable declarations inside rows.Scan loops](https://github.com/krenalis/krenalis/pull/2237/changes/6607f7cffc5151b1713f4a462802214acdd62990).

zapateo (approved):

PR #2239, opened 2026-05-22
all: migrate organization IDs from UUIDs to integers

gazerro:
Update organization identifiers to use integer IDs instead of UUIDs.


gazerro:
See documentation. change: https://github.com/krenalis/krenalis.com/pull/32.

core/internal/state/sendstats.go
```diff
@@ -8,16 +8,16 @@ import (
...
 	"github.com/krenalis/analytics-go"
 )
 
 // sendNotificationStats sends information about notification n to Krenalis.
-func (state *State) sendNotificationStats(client analytics.Client, organization uuid.UUID, n notification) {
+func (state *State) sendNotificationStats(client analytics.Client, organization int, n notification) {
 	go func() {
 		err := client.Enqueue(analytics.Track{
-			UserId: organization.String(),
+			UserId: strconv.Itoa(organization),
```
zapateo, line 20: I'm not entirely convinced by this change to the user ID.

Previously, UUIDs uniquely identified an organization even across multiple installations, whereas now we'd find multiple users all with ID 1.

I'm not sure this is what we want, perhaps we should discuss this for a moment and evaluate our options.

zapateo (changes requested):

gazerro:
See the alternative PR #2249.

PR #2240, opened 2026-05-23
core: replace random internal IDs with database identity columns

gazerro:
```
core: replace random internal IDs with database identity columns

Replace application-generated internal IDs with PostgreSQL identity
columns for workspaces, connections, pipelines, and access keys.

Some internal IDs were already defined as identity columns in the
schema, but values were still generated in application code.

Internal IDs are now assigned sequentially by PostgreSQL instead of
being generated in application code. This removes custom ID generation,
keeps ID ownership in the database, and avoids sparse internal ID
allocation.

To migrate the database and any PostgreSQL data warehouse, start
Krenalis with:

> krenalis -upgrade-db

Snowflake data warehouses are not migrated.
```


gazerro:
See the alternative PR #2249.

PR #2248, opened 2026-05-27
all: introduce enabling and disabling of organizations

zapateo:
For the documentation, see https://github.com/krenalis/krenalis.com/pull/33

```
all: introduce enabling and disabling of organizations

Introduce the concept of "status of an organization", which means that
an organization can be enabled or disabled.

This allows for restrictions based on various criteria (outdated
organizations, organizations exceeding some type of limit, etc.).

More in detail:

* 'core' guarantees that pipelines belonging to a disabled organization
  behave as if they were disabled, returning an OrganizationDisabled
  unprocessable error wherever a PipelineDisabled error would otherwise
  be returned. Event ingestion is the exception: requests authenticated
  with the API key fail with an Unprocessable OrganizationDisabled
  error, while those authenticated with an event write key fail with a
  503 Service Unavailable error.

* Regarding 'cmd', methods used by the Admin console, logins to
  workspaces of a disabled organization, as well as calls made with API
  and MCP keys to its workspaces, return an OrganizationDisabled
  unprocessable error. Calls with Organizations API keys are unaffected.

Please note that, regarding the error returned by the MCP server, we
have a long-standing issue (#1667), so currently the error doesn't have
a specific type but rather generically indicates that the organization
is disabled. Error handling in the MCP needs to be addressed separately.
```


core/core.go
```diff
@@ -430,6 +434,13 @@ func (core *Core) AcceptInvitation(ctx context.Context, token string, name strin
 		if isInvitationTokenExpired(createdAt) {
 			return nil, errors.Unprocessable(InvitationTokenExpired, "invitation token is expired")
 		}
+		organization, ok := core.state.Organization(n.Organization)
```
gazerro: You are not reading the current state of the organization, but a previous version that may be outdated. To ensure that an invitation is not accepted when the organization is disabled, you must read the organization's state directly from the database.

cmd/api.go
```diff
@@ -25,6 +25,9 @@ type api struct {
 // AcceptInvitation accepts the invitation with a given invitation token.
 //
 // Authentication is not required to call AcceptInvitation.
+//
+// If the organization is disabled, an UnprocessableError error with code
```
gazerro: We don't document the errors returned by endpoint handlers. The Krenalis API documentation is the source of truth for the errors they may return.

cmd/api.go
```diff
@@ -294,6 +306,10 @@ func (api api) SendMemberPasswordReset(w http.ResponseWriter, r *http.Request) (
 		return nil, errors.New("there are no organizations")
 	}
 	org := organizations[0]
+	// Verify that the organization is enabled.
```
gazerro: This comment doesn't add anything beyond what is already clearly expressed by the if condition and the error message. The same applies to the other occurrences of this code.

gazerro:
I find it confusing that the organization status is sometimes checked in `cmd` and sometimes in `core`.

The commit message does not clearly explain where this responsibility is supposed to live.

The requirement for `core` should be:

> Pipelines belonging to a disabled organization behave as if they were disabled. In such cases, `core` methods return an `OrganizationDisabled` error.

The requirement for `cmd` should be:

> No operation on a disabled organization is allowed unless it is performed using an Organizations API key.

From what I can see, the following `core` methods currently return an error when the organization is disabled:

* `MemberInvitation`
* `AcceptInvitation`
* `ValidateMemberPasswordResetToken`
* `ChangeMemberPasswordByToken`

I don't think `core` should return an error for these methods. Instead, `cmd` should prevent these operations from being executed when the organization is disabled.

To do this, make `MemberInvitation` and `ValidateMemberPasswordResetToken` return the organization and perform the enabled check in `cmd`.

Before calling `AcceptInvitation` and `ChangeMemberPasswordByToken`, `cmd` should call `MemberInvitation` and `ValidateMemberPasswordResetToken` of `core` respectively and verify that the organization is enabled.

core/internal/state/keep.go
```diff
@@ -462,8 +464,9 @@ func (state *State) createConnection(n notification) string {
 
 // CreateOrganization is the event sent when an organization is created.
 type CreateOrganization struct {
-	ID   string
-	Name string
+	ID      string
+	Name    string
+	Enabled bool
```
gazerro, line 474: I would remove `Enabled`, since it should always be `true`. If it were `false`, it would indicate an inconsistency between the organization's created in the database and its reported state, which would be a bug that should be fixed.

Another way to think about it is that a status change notification should contain all and only the information that cannot be inferred from the state itself.

core/organization.go
```diff
@@ -727,6 +743,29 @@ func (this *Organization) SendMemberPasswordReset(ctx context.Context, email str
...
+	n := state.SetOrganizationStatus{
+		ID:      this.organization.ID,
+		Enabled: enabled,
+	}
+	err := this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
+		result, err := tx.Exec(ctx, "UPDATE organizations SET enabled = $1 WHERE id = $2 AND enabled <> $1", n.Enabled, n.ID)
+		if err != nil {
+			return nil, err
+		}
+		if result.RowsAffected() == 0 {
```
gazerro, line 755: We need to distinguish between a state that has not changed and an organization that does not exist.

core/pipeline.go
```diff
@@ -513,6 +514,9 @@ func (this *Pipeline) Run(ctx context.Context, incremental *bool) (string, error
 			return "", errors.Unprocessable(CannotRunIncrementally, "incremental requires an update time column")
 		}
 	}
+	if !c.Workspace().Organization().Enabled {
+		return "", errors.Unprocessable(OrganizationDisabled, "organization is disabled")
```
gazerro: `"organization %s is disabled"`

core/pipeline.go
```diff
@@ -890,21 +895,27 @@ func (this *Pipeline) createRun(ctx context.Context, incremental *bool) (string,
...
+				"LIMIT 1", n.Pipeline).Scan(&pipelineEnabled, &organizationEnabled, &function, &inc, &cursor, &executing)
 			if err != nil {
 				if err == sql.ErrNoRows {
 					return nil, errors.NotFound("pipeline %s does not exist", n.Pipeline)
 				}
 				return nil, err
 			}
-			if !enabled {
+			if !organizationEnabled {
+				return nil, errors.Unprocessable(OrganizationDisabled, "organization is disabled")
```
gazerro: `"organization %s is disabled"`

core/pipeline_scheduler.go
```diff
@@ -176,6 +177,39 @@ func (ps *pipelineScheduler) onElectLeader(n state.ElectLeader) {
...
+	}
+	if pipelines == nil {
+		return
+	}
+	if n.Enabled {
+		for _, pipeline := range pipelines {
+			ps.executor.AddPipeline(pipeline)
+		}
+	} else {
+		// Unlike other call sites, RemovePipeline is called synchronously to
```
gazerro: This comment is not entirely accurate. The issue is not about avoiding a data race, but rather that a subsequent enable/disable notification could be processed before this one has completed.

It may not even be worth keeping this comment. It would be more appropriate to document the places where the removal is intentionally performed asynchronously.

If you still want to keep a comment here, I think something like the following would be correct:

```go
// Unlike deletion handlers, this listener waits for all removals to                                                                                                                                  
// complete. This preserves notification order: a later re-enabling                                                                                                                                   
// cannot add the pipelines before this disabling has removed them.
```

core/pipeline_scheduler.go
```diff
@@ -176,6 +177,39 @@ func (ps *pipelineScheduler) onElectLeader(n state.ElectLeader) {
...
+	if ps.executor == nil {
+		return
+	}
+	var pipelines []*state.Pipeline
+	org, _ := ps.core.state.Organization(n.ID)
+	for _, workspace := range org.Workspaces() {
+		for _, connection := range workspace.Connections() {
+			for _, pipeline := range connection.Pipelines() {
+				if pipeline.SchedulePeriod != 0 {
+					pipelines = append(pipelines, pipeline)
```
gazerro: I would perform the operation directly rather than postponing it.

core/pipeline_scheduler.go
```diff
@@ -211,7 +245,8 @@ func newPipelineExecutor(core *Core, wg *sync.WaitGroup, ctx context.Context) *p
 		pe.pipelines[i] = map[int16][]*state.Pipeline{}
 	}
 	for _, pipeline := range pe.core.state.Pipelines() {
-		if pipeline.SchedulePeriod == 0 {
+		org := pipeline.Connection().Workspace().Organization()
```
gazerro: Optimize:

```go
if pipeline.SchedulePeriod == 0 {
    continue
}
org := pipeline.Connection().Organization()
if !org.Enabled {
    continue
}
```

core/internal/collector/collector.go
```diff
@@ -107,38 +107,15 @@ func New(db *db.DB, stream streams.Stream, st *state.State, ds *datastore.Datast
 	st.AddListener(c.onDeleteOrganization)
 	st.AddListener(c.onDeleteWorkspace)
 	st.AddListener(c.onLinkConnection)
+	st.AddListener(c.onSetOrganizationStatus)
```
gazerro, line 110: For the collector, a disabled organization is effectively indistinguishable from a non-existent one. When an organization is disabled, `onSetOrganizationStatus` removes all of its workspaces and connections from the collector.

However, the current implementation does not consistently enforce this assumption. For example, `New` adds all organizations to the collector regardless of their status, and `onCreateConnection` does not verify whether the associated organization is enabled.

As a result, the collector's internal state can become inconsistent, leading to unexpected panics. For instance, deleting a disabled organization triggers a panic.

core/core.go
```diff
@@ -695,16 +716,17 @@ func (core *Core) CountOrganizations(ctx context.Context) int {
...
 	if err := util.ValidateStringField("name", name, 45); err != nil {
 		return "", errors.BadRequest("%s", err)
 	}
-	n := state.CreateOrganization{Name: name}
+	n := state.CreateOrganization{Name: name, Enabled: true}
 	for {
 		n.ID = generateID(core.state.Organization)
 		err := core.state.Transaction(ctx, func(tx *dbpkg.Tx) (any, error) {
-			_, err := tx.Exec(ctx, "INSERT INTO organizations (id, name) VALUES ($1, $2)", n.ID, n.Name)
+			_, err := tx.Exec(ctx, "INSERT INTO organizations (id, name, enabled) VALUES ($1, $2, $3)", n.ID, n.Name, n.Enabled)
```
gazerro, line 736: The `enabled` column defaults to `TRUE`, so there is no need to set it explicitly. It would be cleaner to leave it unset.

core/internal/initdb/initdb.go
```diff
@@ -132,7 +132,7 @@ func initialize(ctx context.Context, tx *db.Tx, dockerMember bool) error {
 	}
 	// Insert the organization.
 	organizationID := base58.Generate(12)
-	_, err := tx.Exec(ctx, "INSERT INTO organizations (id, name) VALUES ($1, 'ACME inc')",
+	_, err := tx.Exec(ctx, "INSERT INTO organizations (id, name, enabled) VALUES ($1, 'ACME inc', true)",
```
gazerro, line 135: The `enabled` column defaults to `TRUE`, so there is no need to set it explicitly. It would be cleaner to leave it unset.

core/internal/state/keep.go
```diff
@@ -1281,6 +1285,26 @@ func (state *State) setPipelineSchedulePeriod(n notification) string {
...
+}
+
+// setOrganizationStatus sets the status of an organization.
+func (state *State) setOrganizationStatus(n notification) string {
+	e := SetOrganizationStatus{}
+	if !decodeNotification(n, &e) {
+		return ""
+	}
+	p := state.replaceOrganization(e.ID, func(p *Organization) {
```
gazerro: s/`p`/`o`/

test/krenalistester/support.go
```diff
@@ -181,6 +181,19 @@ func (c *Krenalis) CreateConnection(connection ConnectionToCreate) string {
 	return response.ID
 }
 
+// CreateConnectionErr is like CreateConnection but returns an error instead of
+// failing the test.
+func (c *Krenalis) CreateConnectionErr(connection ConnectionToCreate) (string, error) {
```
gazerro, line 186: I think these new methods introduce an unnecessary amount of code duplication. `CreateConnection` could delegate to `CreateConnectionErr` and panic if an error is returned. The same pattern could be used for the other newly added methods as well.

Additionally, although this would be better handled in a separate pull request, I think `CreateConnection` should be renamed to `MustCreateConnection`, and `CreateConnectionErr` should be renamed to `CreateConnection`. The same applies to the other methods.

test/krenalistester/support.go
```diff
@@ -506,6 +556,14 @@ func (c *Krenalis) RunIdentityResolution() {
 	}
 }
 
+// StartIdentityResolutionErr starts the identity resolution and returns an
+// error instead of failing the test. Unlike [RunIdentityResolution], it does
+// not wait for the identity resolution to complete, so this method is called
+// 'Start', not 'Run'.
+func (c *Krenalis) StartIdentityResolutionErr() error {
```
gazerro, line 564: I find it odd that this method name ends with `Err`. As far as I can tell, the current convention is to use the `Err` suffix for methods that are the error-returning counterpart of another method that fails the test instead of returning an error. However, that does not seem to be the case here.

For example, `TestWorkspaceCreation` also returns an error but does not use the `Err` suffix.

That said, my comment elsewhere still applies: the naming of the `Krenalis` methods should be revisited.

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+)
+
+// TestOrganizationDisabled tests the disable/enable flow of an organization.
+func TestOrganizationDisabled(t *testing.T) {
+
+	// Test's header (copy-paste me in other tests).
+	if testing.Short() {
+		t.Skip()
+	}
+	c := krenalistester.NewKrenalisInstance(t)
```
gazerro, line 27: Any reason for using `c` instead of `k`?

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+	c.Start()
+	defer c.Stop()
+
+	// Retrieve the organization ID.
+	orgs := c.Organizations(0, 100)
+	if len(orgs) != 1 {
+		t.Fatalf("expected exactly one organization, got %d", len(orgs))
+	}
+	orgID := orgs[0].ID
```
gazerro, line 36: Better:  `org := orgs[0]`

You can write: `if !org.Enabled { ... }`

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// Test that the organization is enabled at startup.
+	if !orgs[0].Enabled {
+		t.Fatal("organization should be enabled, but it is disabled")
+	}
+
+	// Verify that the endpoint for changing the organization state requires the
+	// organization API.
+	t.Run("set status without organizations API key is rejected", func(t *testing.T) {
```
gazerro, line 45: This test does not do what it claims to do.

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+		},
+	})
+	run := c.RunPipeline(importPipeline)
+	c.WaitRunsCompletion(dummySrc, run)
+
+	// Set up a JavaScript source with an Event pipeline so that events can be
+	// ingested through the /v1/events endpoint. This must be done while the
+	// organization is still enabled, as creating connections and pipelines is
+	// itself rejected once the organization is disabled.
+	ctx := context.Background()
```
gazerro: This context is created here, but it is not used until much later.

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+		t.Fatal("expected an error, got nil")
+	}
+	statusErr, ok := err.(*krenalistester.StatusCodeError)
+	if !ok {
+		t.Fatalf("expected *StatusCodeError, got %T: %v", err, err)
+	}
+	if statusErr.Code != http.StatusUnprocessableEntity {
+		t.Fatalf("expected HTTP status %d, got %d: %s", http.StatusUnprocessableEntity, statusErr.Code, statusErr.ResponseText)
+	}
+	if !strings.Contains(statusErr.ResponseText, `"code":"OrganizationDisabled"`) {
```
gazerro: The test is weak. It should decode the error into a struct before performing the assertions.

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+		t.Fatalf("expected error code %q in response, got: %s", "OrganizationDisabled", statusErr.ResponseText)
```
gazerro: I do not understand why `OrganizationDisabled` is quoted while the unexpected error is not. If anything, I would expect the opposite.

Also, the colon does not belong here. It is not consistent with how we usually format this kind of error message.

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+	t.Helper()
+	if err == nil {
+		t.Fatal("expected an error, got nil")
+	}
+	statusErr, ok := err.(*krenalistester.StatusCodeError)
+	if !ok {
+		t.Fatalf("expected *StatusCodeError, got %T: %v", err, err)
+	}
+	if statusErr.Code != http.StatusUnprocessableEntity {
+		t.Fatalf("expected HTTP status %d, got %d: %s", http.StatusUnprocessableEntity, statusErr.Code, statusErr.ResponseText)
```
gazerro, line 330: `"expected HTTP status %d, got %d (%q)"`

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+	// Verify that the endpoint for changing the organization state requires the
+	// organization API.
+	t.Run("set status without organizations API key is rejected", func(t *testing.T) {
+		err := c.SetOrganizationStatusErr(orgID, false, http.Header{"Krenalis-Workspace": nil})
+		statusErr, ok := err.(*krenalistester.StatusCodeError)
+		if !ok {
+			t.Fatalf("expected *StatusCodeError, got %T: %v", err, err)
+		}
+		if statusErr.Code != http.StatusUnauthorized {
+			t.Fatalf("expected HTTP status %d, got %d: %s", http.StatusUnauthorized, statusErr.Code, statusErr.ResponseText)
```
gazerro, line 52: `"expected HTTP status %d, got %d (%q)"`

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+		_, err := c.CreateConnectionErr(krenalistester.ConnectionToCreate{
+			Name:      "Dummy that should not be created",
+			Role:      krenalistester.Source,
+			Connector: "dummy",
+			Settings:  json.Value("{}"),
+		})
+		assertOrganizationDisabled(t, err)
+	})
+
+	t.Run("create pipeline is rejected", func(t *testing.T) {
```
gazerro, line 129: I am not sure how useful it is to test some API calls but not others. At most, I would expect the API calls to be grouped according to some criterion, with a single test covering each group.

test/organization_disabled_test.go
```diff
@@ -0,0 +1,312 @@
...
+		err := c.Call("POST", "/v1/events/track",
+			http.Header{"Authorization": []string{"Bearer " + writeKey}},
+			map[string]any{
+				"userId": "dropped-while-disabled-typed",
+				"event":  "Typed event sent while disabled",
+			}, nil)
+		if err != nil {
+			t.Fatalf("POST /v1/events/track must not be rejected while the organization is disabled, got: %v", err)
+		}
+		time.Sleep(2 * time.Second)
```
gazerro, line 253: This test assumes that, if the events were collected instead of discarded, they would be written to the warehouse within two seconds. If they were written later, the test would incorrectly pass.

Since no waiting period can guarantee that the events will never be written to the warehouse, I think it would be better to start this check as early as possible and verify it at the end of the test. That would reduce the chance of false positives.

core/core.go
```diff
@@ -695,16 +716,17 @@ func (core *Core) CountOrganizations(ctx context.Context) int {
 
 // CreateOrganization creates a new organization and returns its identifier.
 // name cannot be empty and cannot be longer than 45 runes.
+// The created organization is already enabled.
```
gazerro: There is currently an inconsistency in the API defaults: omitting the `enabled` field when creating a pipeline results in a disabled pipeline, whereas omitting it when creating an organization results in an enabled organization (and in fact the field cannot currently be specified for organizations at all).

Additionally, if we decide to introduce an `enabled` field for organizations in the future, its default value should be `true`, not `false`, to preserve backward compatibility.

cmd/organization.go
```diff
@@ -243,6 +243,29 @@ func (organization organization) Members(_ http.ResponseWriter, r *http.Request)
...
+// SetStatus sets the status of an organization.
+func (organization organization) SetStatus(w http.ResponseWriter, r *http.Request) (any, error) {
+	if err := organization.authenticateOrganizationsRequest(r); err != nil {
+		return nil, err
+	}
+	if err := validateRequiredBody(w, r, false); err != nil {
+		return nil, err
+	}
+	var body struct {
+		Enabled bool `json:"enabled"`
```
gazerro, line 255: When the request body is `{}`, the method silently sets `enabled` to `false` rather than returning an error.

core/pipeline.go
gazerro: If the organization is disabled, the method should return `OrganizationDisabled` instead of `CannotRunIncrementally`.

core/internal/collector/collector.go
gazerro, line 893: Should this call fail if the organization is disabled? Is it expected for the settings of a disabled organization to be returned?

core/pipeline_scheduler.go
```diff
@@ -53,6 +53,7 @@ func newPipelineScheduler(core *Core) *pipelineScheduler {
 	core.state.AddListener(ps.onDeletePipeline)
 	core.state.AddListener(ps.onDeleteWorkspace)
 	core.state.AddListener(ps.onElectLeader)
+	core.state.AddListener(ps.onSetOrganizationStatus)
```
gazerro, line 60: The expected behavior of the scheduler with disabled organizations is unclear. `onSetOrganizationStatus` suggests that disabled organizations should be excluded from the scheduler entirely, but the rest of the implementation does not appear to follow the same rule consistently.

gazerro (changes requested):

zapateo:
> I find it confusing that the organization status is sometimes checked in `cmd` and sometimes in `core`.
>  [...]

The new revised specification is this:

* **core** guarantees that pipelines belonging to a disabled organization
  behave as if they were disabled, returning an OrganizationDisabled
  unprocessable error wherever a PipelineDisabled error would otherwise
  be returned. Event ingestion is the exception: requests authenticated
  with the API key fail with an Unprocessable OrganizationDisabled
  error, while those authenticated with an event write key fail with a
  503 Service Unavailable error.

* Regarding **cmd**, methods used by the Admin console, logins to
  workspaces of a disabled organization, as well as calls made with API
  and MCP keys to its workspaces, return an OrganizationDisabled
  unprocessable error. Calls with Organizations API keys are unaffected.

These are the commits that go in that direction:

1. [Stop core from removing MCP server for disabled organizations](https://github.com/krenalis/krenalis/pull/2248/commits/b9a9bc2a5c2d3cf67c9b953a4868f95327b0f29c)
2. [Update and simplify comment of core.Organization.Enabled](https://github.com/krenalis/krenalis/pull/2248/commits/3084fdbbc6f81750c31e45ae8ae18639890b7bdd). Changed by a following commit.
3. [Refactor member and invitation methods (updated commit)](https://github.com/krenalis/krenalis/pull/2248/commits/223486b46515e05816770303bce3485363ce8b05)
4. [Update comment of core/organization.Enabled](https://github.com/krenalis/krenalis/pull/2248/commits/d4e208d8240dd4b7ed3673f437299b7456b46b30)
5. [Change events ingestion when organization is disabled](https://github.com/krenalis/krenalis/pull/2248/commits/d0825482e64f29ffbca34fa26d1973d202ef2f23)
6. [Remove unused local variable organizationID](https://github.com/krenalis/krenalis/pull/2248/commits/32e0167e5ef1b4aa561dea172422b539ed3e8661)

cmd/api.go
zapateo: Done in:

1. [Refactor member and invitation methods (updated commit)](https://github.com/krenalis/krenalis/pull/2248/commits/223486b46515e05816770303bce3485363ce8b05)
2. [Remove doc about OrganizationDisabled from SendMemberPasswordReset](https://github.com/krenalis/krenalis/pull/2248/commits/f7a1c69b2d5f56832bfee17a3485e85e341e6030)
3. [Do not document UnprocessableErrors in apis_server.go](https://github.com/krenalis/krenalis/pull/2248/commits/ec8420cc079e101cdde5cf71f57db9532961e757)

core/core.go
zapateo: With the changes made, I believe this conversation no longer applies, or am I wrong?

cmd/api.go
zapateo: See [Remove unnecessary comments // Verify that the organization is ...](https://github.com/krenalis/krenalis/pull/2248/commits/2ed20e2aad37800c37e18567bd3dfeea7b5ff96b).

Note that I left the comment in `authenticateAdminRequest` for consistency, as in that method we document almost each code block, and removing the comment would make – in my opinion – the code unclear.

core/internal/state/keep.go
zapateo: Done in [Remove "Enabled" from org creation status notification](https://github.com/krenalis/krenalis/pull/2248/commits/2d74547f9832d83dd5debaf66af0fd98bfd5fe79).

EDIT: after the discussion, this has been changed again in [Add the "Enabled" field when creating organizations](https://github.com/krenalis/krenalis/pull/2248/commits/ecc5d72fac8289d602815c9de10ac70e2c384349).

core/pipeline.go
zapateo: Done in [Add organization ID to "organization is disabled" errors](https://github.com/krenalis/krenalis/pull/2248/commits/a3eea517daa84d5e7c0ad9fc259e498eb5636120).

core/pipeline.go
zapateo: Done in [Add organization ID to "organization is disabled" errors](https://github.com/krenalis/krenalis/pull/2248/commits/a3eea517daa84d5e7c0ad9fc259e498eb5636120).

core/core.go
zapateo: Done in [Rely on default for "enabled" in SQL queries when creating organizations](https://github.com/krenalis/krenalis/pull/2248/commits/4a73a0ecd682b504f02228742ac49572fd09bf06).

EDIT: after the discussion, this has been changed again in [Add the "Enabled" field when creating organizations](https://github.com/krenalis/krenalis/pull/2248/commits/ecc5d72fac8289d602815c9de10ac70e2c384349).

core/internal/initdb/initdb.go
zapateo: Done in [Rely on default for "enabled" in SQL queries when creating organizations](https://github.com/krenalis/krenalis/pull/2248/commits/4a73a0ecd682b504f02228742ac49572fd09bf06).

EDIT: after the discussion, this has been changed again in [Add the "Enabled" field when creating organizations](https://github.com/krenalis/krenalis/pull/2248/commits/ecc5d72fac8289d602815c9de10ac70e2c384349).

core/internal/state/keep.go
zapateo: Done in [Rename local variable p to o in setOrganizationStatus](https://github.com/krenalis/krenalis/pull/2248/commits/dbeea0d37b3e17b370080514bcadd167db4a7f94).

test/organization_disabled_test.go
zapateo: We use the variable `c` in all tests; it's a holdover from the "Chichi" days.

In general, it's best to keep the test header (the one marked with the comment `// Test's header (copy-paste me in other tests).`) as similar as possible across tests.

So, if you agree, I'll make this rename later (`c` → `k`) in `main`, when this PR is merged. What do you think?

test/organization_disabled_test.go
zapateo: I think this would make the test more confusing.

Assuming that every time we call `c.Organization(orgID)`, we actually call the API endpoint to return the new organization, this is part of the test.

As the code stands now, the "fixed point" is the organization ID (stored in `orgID`), which is retrieved initially and does not change, and throughout the rest of the test, we want to re-retrieve the organization.

With your proposed change, we could avoid the first call (having just retrieved the organization), which is effectively useless for the test, but I think the change in variable names would create confusion later on.

cmd/organization.go
zapateo: As discussed, I've opened the issue https://github.com/krenalis/krenalis/issues/2279.

core/core.go
zapateo: As discussed, changed in [Add the "Enabled" field when creating organizations](https://github.com/krenalis/krenalis/pull/2248/commits/ecc5d72fac8289d602815c9de10ac70e2c384349).

core/pipeline_scheduler.go
zapateo: Done in [Optimize check in newPipelineExecutor](https://github.com/krenalis/krenalis/pull/2248/commits/cbb2a0ddbce1fff042c69d606a61cd900382c204).

test/krenalistester/support.go
zapateo: I agree that the names of the various test methods could be revised.

In this particular case, as documented, conceptually the `StartIdentityResolutionErr` method is somewhat of a counterpart to the `RunIdentityResolution` method, but the former returns an error if it fails to start (not to run), so conceptually the name should have been different.

If you have a better idea in this case, let me know; otherwise, I propose opening an issue and reviewing these names as a whole on the `main` branch once this PR has been merged.

test/organization_disabled_test.go
zapateo: Is this better? Or did you mean something else? [Change unauthorized set status test documentation](https://github.com/krenalis/krenalis/pull/2248/commits/269d053d8252d1c99d60aa168b803b54a0d0d8dc)

test/organization_disabled_test.go
zapateo: See the commit [Use T.Context instead of context.Background](https://github.com/krenalis/krenalis/pull/2248/commits/50ffbbd81d2ed112f2ff6589f2da67cf6717f637).

core/pipeline_scheduler.go
zapateo: As discussed, I merged the two comments. See [Update the comment about race in onSetOrganizationStatus](https://github.com/krenalis/krenalis/pull/2248/commits/905b88a6179ef080b66a523aca11ad50ba075771).

test/organization_disabled_test.go
zapateo: As discussed verbally, see [Improve events import test checks and timing](https://github.com/krenalis/krenalis/pull/2248/commits/9f0271fb4a17a10481e7dc3f089a33c7cb8ffd19).

test/organization_disabled_test.go
zapateo: I don't have a specific criterion in mind for grouping various methods to test.

The idea here was to add a few scattered tests, just to increase the chance of catching any errors. In general, all the Go tests in the '/test' directory follow this principle: they're not unit tests, so they don't exhaustively test anything, nor are they limited to testing just one specific feature; we simply add additional tests when possible, which can fail even if a problem is introduced in a part of the software not strictly related to the one being tested.

One reason not to add tests of this type is the cost in terms of time and resources, but these calls are very fast, and I thought having more, even scattered, was better than having none.

In light of this, what do you think?

core/organization.go
zapateo: As discussed verbally, this should be correct, and it's the same thing we do when managing a pipeline's state change (which I used as a reference for the organizations code).

It seems to be correct because, if the organization didn't exist, the error would have already been returned earlier and we wouldn't have gotten to this point. However, if it had been deleted in the meantime, the user would be indifferent whether they receive a "not found" or an "ok, done".

core/pipeline_scheduler.go
zapateo: Done in [core: perform operations directly in onSetOrganizationStatus](https://github.com/krenalis/krenalis/pull/2248/commits/d7d1fca2d5629b9c1ce93685a0992015e23fa905).

test/organization_disabled_test.go
zapateo: I used the same style we already use here:

https://github.com/krenalis/krenalis/blob/6fc3fb5f695fbcc4b665cbbd144f857df8b12a96/test/workspace_scoped_api_key_test.go#L54-L69

which was introduced by this commit: https://github.com/krenalis/krenalis/commit/6fc3fb5f695fbcc4b665cbbd144f857df8b12a96

If you'd rather change it, maybe I'd change it consistently there too in a separate change, tell me what you think is best.

test/organization_disabled_test.go
zapateo: See [the other comment](https://github.com/krenalis/krenalis/pull/2248#discussion_r3428996659).

core/pipeline.go
zapateo: We can (**solution A**) move the organization enablement check before the incremental check. In that case, the organization disabled error would be returned first (good), but we would have an Unprocessable error returned before a Bad Request (not sure if it is good).

Otherwise, we can (**solution B**):

1. Formally check incrementally, possibly returning a Bad Request
2. Check organization enablement, possibly returning an Unprocessable
3. Perform the other incremental check, possibly returning an Unprocessable

The solution A is simpler; if it's not a problem (perhaps we already "mix" Unprocessable and Bad Request errors elsewhere), we can implement that.

Let me know what you think is better. Or do you see other alternatives?

test/organization_disabled_test.go
zapateo: I updated the test in the commit [Update test by decoding into struct and doing more strict checks](https://github.com/krenalis/krenalis/pull/2248/commits/185d33019b95121ae129c1e34c24f042ecfba8ca).

test/organization_disabled_test.go
zapateo: I updated the text in the commit [Update test by decoding into struct and doing more strict checks](https://github.com/krenalis/krenalis/pull/2248/commits/185d33019b95121ae129c1e34c24f042ecfba8ca).

core/internal/collector/collector.go
zapateo: As discussed, we do not expect this method to return an error.

test/krenalistester/support.go
zapateo: Yes, I agree. As [the other comment where I replied](https://github.com/krenalis/krenalis/pull/2248#discussion_r3427016650), I'd rather leave this branch as is and review everything as a whole later on `main`.

core/internal/collector/collector.go
zapateo: See these commits here. They don't just address the collector specifically, but generally address how a component should treat a disabled organization as nonexistent.

Do these changes seem correct to you? Do you see any other issues about this?

1. [Fix collector.New regarding disabled organizations](https://github.com/krenalis/krenalis/pull/2248/commits/e89b92eee1952ce527ebc98c1733e4e41875ae65)
2. [Do not process deletion of organizations that were disabled](https://github.com/krenalis/krenalis/pull/2248/commits/fddc6b446496b5cc2fce71e912033de0f0e8c864)
3. [Do not handle disabled organizations in newDestinations](https://github.com/krenalis/krenalis/pull/2248/commits/fbae46021ac3d7f32ca3b6aec28b8b431daca232)
4. [Handle disable orgs in every notification listener](https://github.com/krenalis/krenalis/pull/2248/commits/acf78fd5c673caf726393ae8fc1ac4d149257cdb)
5. [Change checks order for consistency with other methods](https://github.com/krenalis/krenalis/pull/2248/commits/a9006421494ac60e3ae6858757dad6412b6212d6)

core/pipeline_scheduler.go
zapateo: See these commits:

1. [Introduce method Pipeline.Organization()](https://github.com/krenalis/krenalis/pull/2248/commits/c7155bfd40ebd95be7bbd625feeed7d3ba2a24ee)
2. [Handle disable orgs in every notification listener](https://github.com/krenalis/krenalis/pull/2248/commits/acf78fd5c673caf726393ae8fc1ac4d149257cdb). This commit, as discussed, is necessary because _core_ makes no guarantees that state change notifications will not arrive (certain operations are restricted at the _cmd_ level, but not at the _core_ level), so, to comply with the principle that disabled orgs are treated (in the three components modified by the commit) as non-existent, the case of disabled orgs must be handled in all the various listener methods.

core/pipeline.go
gazerro: **Solution B**. It's possible that we have other cases where an Unprocessable response is returned before a Bad Request response, but I think there should be a very strong reason for doing so. In this case, I don't feel the reason is strong enough.

core/internal/state/keep.go
```diff
@@ -235,6 +235,7 @@ func (state *State) replaceConnection(id string, f func(*Connection)) *Connectio
 	for _, pipeline := range c.pipelines {
 		pipeline.mu.Lock()
 		pipeline.connection = cc
+		pipeline.organization = cc.organization
```
gazerro: This method must replace all occurrences of `c` with `cc` in the state; therefore, the organization must not be changed.

core/internal/collector/collector.go
```diff
@@ -705,14 +705,19 @@ func (c *Collector) serveEvents(w http.ResponseWriter, r *http.Request) error {
...
+			}
 		} else {
 			// Authenticate with the event write key in the header.
 			connection, _ = c.connectionByKey(token)
 			if connection == nil {
 				return errors.Unauthorized("event write key in the Authorization header is not valid")
 			}
 			usingWriteKey = true
+			if org := connection.Organization(); !org.Enabled {
+				return errors.Unavailable("organization %s is disabled", org.ID)
```
gazerro: When authentication is performed using an event write token, I wouldn't include the organization ID in the error message. I also wouldn't include it when the token is passed in the request body.

test/organization_disabled_test.go
gazerro: Whatever works best for you.

test/organization_disabled_test.go
gazerro: Feel free to leave it as it is.

test/organization_disabled_test.go
zapateo: Personally, I don't see any advantage in making this change in the various places where it is needed, so I would tend to leave it as is.

core/pipeline.go
zapateo: Done in [Change incremental checks](https://github.com/krenalis/krenalis/pull/2248/commits/550f06bba19e5afe2cb5bdd074f65cf32e7aaa1a).

core/internal/collector/collector.go
zapateo: Done in [Remove org ID from unavailable errors](https://github.com/krenalis/krenalis/pull/2248/commits/58785f8975a3dd9a344df6482045277a85fad05d).

core/internal/state/keep.go
zapateo: Done in [Don't change organization in replaceConnection](https://github.com/krenalis/krenalis/pull/2248/commits/ac97acf0f0c3ae9e486814b8eeeae127ce010ebc).

gazerro (approved):

PR #2249, opened 2026-05-27
all: adopt Base58 resource identifiers

gazerro:
```
all: adopt Base58 resource identifiers

Replace the existing identifier types for organizations, access keys,
members, workspaces, connections, pipelines, and pipeline runs with
random 12-character Bitcoin Base58 identifiers, such as 4fr7Kp9ZaQ2m.

The goal is to use a single identifier format that is compact,
URL-friendly, safe to expose through the API, and large enough to make
collisions practically negligible for the expected scale of the system.

The new identifiers provide approximately 70.3 bits of entropy. This is
lower than the 122 bits provided by the previous organization
identifiers, but significantly higher than the 31 bits provided by the
previous integer-based identifiers used by the other resources.

Bitcoin Base58 was chosen instead of Base64 because it avoids two kinds
of problematic characters.

First, it excludes visually ambiguous characters, such as 0, O, I, and
l, which can be hard to distinguish in some fonts.

Second, it excludes characters that are less convenient in URLs, such as
+ and /, which are used by Base64.

This makes identifiers easier to read, copy, paste, and use safely in
URL paths.

Bitcoin Base58 was also chosen instead of extending the existing integer
identifiers from 32 to 64 bits for two main reasons.

First, it keeps URLs shorter. When an identifier is used in a URL path,
a 12-character Base58 value is shorter than a 64-bit integer represented
with up to 19 decimal digits.

Second, it avoids JSON interoperability issues with large integers. Many
JSON libraries and runtimes parse numeric values as float64, which
cannot safely represent all 64-bit integers. Serializing 64-bit integers
as strings would avoid the precision issue, but at that point the
identifier would no longer behave as a numeric value across the API.

This design also leaves room for future optimization. If database
performance becomes a concern, Krenalis can introduce internal 64-bit
integer IDs later while keeping these Base58 identifiers as stable
external API identifiers. That would preserve API compatibility while
allowing the database schema to evolve.

This internal/external ID split was not introduced now because it would
add complexity without a current need. Maintaining two identifiers for
the same resource would make the data model less direct and make
debugging less convenient.
```


gazerro:
See documentation: https://github.com/krenalis/krenalis.com/pull/35

cmd/apis_server.go
```diff
@@ -461,27 +459,12 @@ func (s *apisServer) secureCookie(ctx context.Context) (*securecookie.SecureCook
...
-			return 0, false
-		}
-		n = n*10 + int64(c-'0')
-		if n > math.MaxInt32 {
-			return 0, false
-		}
-	}
-	if n < 1 {
-		return 0, false
+// parseID parses a public row identifier.
```
zapateo, line 460: What is meant here by:

> public row identifier

?

It makes me think of parsing a database identifier, I don't know if the meaning here is intentional or not.

core/core.go
```diff
@@ -694,22 +795,38 @@ func (core *Core) CountOrganizations(ctx context.Context) int {
...
-		return uuid.Nil, errors.BadRequest("%s", err)
+		return "", errors.BadRequest("%s", err)
 	}
 	n := state.CreateOrganization{Name: name}
-	err := core.state.Transaction(ctx, func(tx *dbpkg.Tx) (any, error) {
-		err := tx.QueryRow(ctx, "INSERT INTO organizations (name) VALUES ($1) RETURNING id", name).Scan(&n.ID)
+	for {
+		n.ID = generateID(core.state.Organization)
+		var err error
+		n.GlobalID, err = uuid.NewRandom()
```
zapateo: The concept of “Global ID” for organizations is introduced at various points in this PR, but I don't see it mentioned in the commit message or in the PR on krenalis.com.

I'd like to better understand its purpose and why it was necessary, and then consider documenting it somewhere. If you'd like, we can discuss it face-to-face.

core/core.go
```diff
@@ -442,7 +543,7 @@ func (core *Core) AcceptInvitation(ctx context.Context, token string, name strin
 // AccessKey returns the organization and workspace identifiers associated with
 // the provided access key token and type. If the access key is not restricted
 // to a workspace, the workspace identifier will be 0.
```
zapateo: > the workspace identifier will be 0

This should be updated.

core/internal/connections/httpclient/client.go
```diff
@@ -74,9 +74,9 @@ type endpointGroup struct {
 type Client struct {
 	http           *HTTP
 	connector      string // connector code
-	connection     int    // connection identifier; it is 0 if the client is not relative to a connection
-	clientSecret   string // client secret; only if connection == 0
-	accessToken    string // access token; only if connection == 0
+	connection     string // connection identifier; it is 0 if the client is not relative to a connection
```
zapateo: > it is 0 if the client is not relative to a connection

This should be updated.

core/core.go
```diff
@@ -1756,6 +1872,12 @@ func EventSchema() types.Type {
 	return schemas.Event
 }
 
+// IsValidID reports whether id is a valid identifier for members, access keys,
+// workspaces, connections, pipelines, and pipeline runs.
+func IsValidID(id string) bool {
```
zapateo, line 1868: The comment says:

> // IsValidID reports whether id is a valid identifier for members, access keys,
> // workspaces, connections, pipelines, and pipeline runs.

But this is also used for organizations, isn't it?

Otherwise we can avoid listing all the various elements and use a more generic term (like "resource"), since listing all the various element may not add much, but an out-of-sync comment can be misleading.

What do you think?

core/organization.go
```diff
@@ -239,113 +250,119 @@ func (this *Organization) AccessKeys(ctx context.Context) ([]*AccessKey, error)
...
 	this.core.mustBeOpen()
+
 	if err := util.ValidateStringField("name", name, 100); err != nil {
-		return 0, "", errors.BadRequest("%s", err)
+		return "", "", errors.BadRequest("%s", err)
 	}
-	if workspace < 0 || workspace > maxInt32 {
-		return 0, "", errors.BadRequest("workspace is not a valid workspace identifier")
+	if workspace != "" && !IsValidID(workspace) {
+		return "", "", errors.BadRequest("workspace is not a valid workspace identifier")
```
zapateo: Elsewhere, if an invalid identifier is present, we include it in error messages using the `%q` verb.

(you can search for `%q is not a valid` in the code)

Here, however, we're not inserting the workspace identifier. This was already the case on `main`, but I think it might be worth changing it here in this branch, since it's a change related to identifiers.

We do the same thing here:

https://github.com/krenalis/krenalis/blob/1db3bb72e8ff94e9fea608e3184dd4238e1b09ea/cmd/organization.go#L108

What do you think?

core/connection.go
```diff
@@ -48,14 +48,14 @@ type Connection struct {
...
-	ID                int           `json:"id"`
+	ID                string        `json:"id"`
 	Name              string        `json:"name"`
 	Connector         string        `json:"connector"`
 	ConnectorType     ConnectorType `json:"connectorType"`
 	Role              Role          `json:"role"`
 	Strategy          *Strategy     `json:"strategy"`
 	SendingMode       *SendingMode  `json:"sendingMode"`
-	LinkedConnections []int         `json:"linkedConnections,omitempty"`
+	LinkedConnections []string      `json:"linkedConnections,omitempty"`
```
zapateo, line 58: In the documentation on krenalis.com, this should be updated:

https://github.com/krenalis/krenalis.com/blob/f5f906d9864d5bc0d63a2584176e3dfa0e501d57/spec/connections.go#L55

zapateo (changes requested):
Please, also resolve the merge conflicts.

retini (approved):

cmd/apis_server.go
gazerro: As discussed, I've removed the `parseId` function: [Remove the parseID function and simplify the code](https://github.com/krenalis/krenalis/pull/2249/changes/e9686476b32c78a8bcf593fdac0aab15419ffcf0)

core/core.go
gazerro: Fixed in commit [Fix comments](https://github.com/krenalis/krenalis/pull/2249/changes/6f85519347cc6b3da3143ecf5a25dbe7672336ef).

core/internal/connections/httpclient/client.go
gazerro: Fixed in commit [Fix comments](https://github.com/krenalis/krenalis/pull/2249/changes/6f85519347cc6b3da3143ecf5a25dbe7672336ef).

core/core.go
gazerro: Fixed in commit [Fix `IsValid' comment](https://github.com/krenalis/krenalis/pull/2249/changes/18cbb072f866bfcecc168d5aa1ae630ad31328bd).

I preferred to explicitly list the resources that use this ID format, but I revised the wording to make it more readable.

core/organization.go
gazerro: Done in commit [Add invalid workspace identifier to error messages](https://github.com/krenalis/krenalis/pull/2249/changes/b0026c6fb495d77ae3d9f18360e45687c8087140).

core/connection.go
gazerro: Fixed in commit [Fix linkedConnections](https://github.com/krenalis/krenalis.com/pull/35/changes/7b7a74d07876893e52d1d6a465ddf3643b1b483d).

core/core.go
gazerro: As discussed, I've removed the Global ID, see commits [Remove organization Global ID](https://github.com/krenalis/krenalis/pull/2249/changes/3ee1b0cb02deb659f16bd923a26e12c9fa00baaf) and [Reduce changes to keep.go to the bare minimum](https://github.com/krenalis/krenalis/pull/2249/changes/e9f67ff2b0f7d906e4e468121b0e9bc2753a257f).

gazerro:
> See documentation: [krenalis/krenalis.com#35](https://github.com/krenalis/krenalis.com/pull/35)

Done in commit [Merge branch 'main' into adopt-base58-ids](https://github.com/krenalis/krenalis/pull/2249/changes/7fbaaf4da8925ad32ab886ddc0bcc2aac507781f).

core/internal/initdb/initdb.go
```diff
@@ -131,36 +128,22 @@ func initialize(ctx context.Context, tx *db.Tx) error {
...
-// by the Admin).
-//
-// This function is intended to be called after a successful call to initialize,
-// on its same transaction.
-//
-// Specifically, this function:
-//
-//  1. Deletes the members already present in the PostgreSQL database;
-//
-//  2. Creates a new member whose email is "docker@krenalis.com" and whose
```
zapateo, line 150: This comment, which indicated the cleartext password (`krenalis-password`), has been removed by this PR, so only the hashed password remains in the source, which in fact cannot be used.

In my opinion, the reference to the cleartext password should be re-added.

core/internal/initdb/initdb.go
```diff
@@ -131,36 +128,22 @@ func initialize(ctx context.Context, tx *db.Tx) error {
...
-	if err != nil {
-		return err
+	// Insert the member.
+	memberID := base58.Generate(12)
+	email := "acme@krenalis.com"
+	if dockerMember {
+		email = "docker@krenalis.com"
 	}
+	_, err = tx.Exec(ctx, "INSERT INTO members (id, organization, name, email, password, created_at)\n"+
+		"VALUES ($1, $2, 'User', $3, '$2a$10$1arUoJQAeIVLAuNiErG29ex2r43n/4bJZWmW/PPOiWaSt4ZCH5Ysm', now() at time zone 'utc')",
```
zapateo: On `main`, the default (non-Docker) user name is `ACME inc`:

https://github.com/krenalis/krenalis/blob/6d5ff1552939a049f2224ed569f1702da6f43f10/core/internal/initdb/schema.sql#L43-L46

With this change, the name of this user becomes `User`:

https://github.com/krenalis/krenalis/blob/3ee1b0cb02deb659f16bd923a26e12c9fa00baaf/core/internal/initdb/initdb.go#L138-L146

If this is intentional, that's fine with me. In that case, since it is not strictly related to this change, I would document it in the commit message as an additional change done by this PR.

core/internal/initdb/upgrade_base58.go
```diff
@@ -0,0 +1,551 @@
...
+		`ALTER TABLE pipelines_errors ADD CONSTRAINT pipelines_errors_pipeline_fkey FOREIGN KEY (pipeline) REFERENCES pipelines ON DELETE CASCADE`,
+		`ALTER TABLE pipelines_metrics ADD CONSTRAINT pipelines_metrics_pipeline_fkey FOREIGN KEY (pipeline) REFERENCES pipelines ON DELETE CASCADE`,
+		`CREATE INDEX ON pipelines_errors (pipeline)`,
+		`CREATE INDEX ON pipelines_metrics (pipeline)`,
+		`TRUNCATE notifications`,
+	}
+}
+
+func addBase58Check(table, column string) string {
+	name := table + "_" + column + "_base58_check"
```
zapateo: If I understand correctly, the name of the constrain after a database upgrade (`organizations_id_base58_check`) is different from the name of the constrain resulting from a clean install (`organizations_id_check`).

I don't know how significant this discrepancy is, or whether we can ignore it or not, but I'm thinking about the possibility that, in the future, we want to modify this constrain and find ourselves having to manage both names.

For reproducing this, on any PostgreSQL database, running these SQL statements simulates both a clean install and an upgrade. Note that **these delete the `organizations` table, if it exists in the database they're run on**.

```sql
-- case #1: fresh table.
DROP TABLE IF EXISTS organizations;
CREATE TABLE organizations (
    id varchar(12) NOT NULL CHECK (id ~ '^[1-9A-HJ-NP-Za-km-z]{12}$'),
    PRIMARY KEY (id)
);
SELECT n.nspname AS schema, c.conname AS constraint_name
    FROM pg_constraint c
    JOIN pg_namespace n ON n.oid = c.connamespace
    WHERE c.contype = 'c'
    ORDER BY schema;

-- case #2: "migrated" table.
DROP TABLE IF EXISTS organizations;
CREATE TABLE organizations (
    id varchar(12) NOT NULL,
    PRIMARY KEY (id)
);
ALTER TABLE organizations
    ADD CONSTRAINT organizations_id_base58_check
    CHECK (id ~ '^[1-9A-HJ-NP-Za-km-z]{12}$');
SELECT n.nspname AS schema, c.conname AS constraint_name
    FROM pg_constraint c
    JOIN pg_namespace n ON n.oid = c.connamespace
    WHERE c.contype = 'c'
    ORDER BY schema;
```

I wanted to bring this to our attention, assuming my SQL code actually mimics Krenalis's situation.

In that case, let's consider whether it's worth standardizing or not. What do you think?

warehouses/snowflake/quote.go
```diff
@@ -108,3 +108,17 @@ func quoteString(b *strings.Builder, s string) {
 	}
 	b.WriteByte('\'')
 }
+
+// quoteStringForDynamicSQL quotes s as a SQL string literal and escapes the
+// quote delimiters for embedding in a single-quoted dynamic SQL body.
+func quoteStringForDynamicSQL(b *strings.Builder, s string) {
+	if strings.IndexAny(s, "\x00'\b\f\n\r\t\\") == -1 {
```
zapateo: Here we could use [`strings.ContainsAny`](https://pkg.go.dev/strings#ContainsAny).

Decide for yourself whether it's worth it or not.

core/internal/initdb/upgrade_base58.go
zapateo, line 1: To test the update procedure, I performed these steps:

1. I initialized a Krenalis DB and warehouse on `main`
2. I switched to this branch
3. I ran `go run -tags dev . -upgrade-db`

I get this error:
```
2026/06/04 10:18:53 INFO PostgreSQL database Base58 ID migration prepared successfully
error: cannot migrate warehouse for workspace ZTM1t1Ymj2fo: ERROR: cannot drop column connection_id of table krenalis_events because other objects depend on it (SQLSTATE 2BP01)
exit status 1
```

Am I doing something wrong?

zapateo (changes requested):

core/internal/initdb/initdb.go
gazerro: Done in commit [Restore the plain-text password in initdb comments](https://github.com/krenalis/krenalis/pull/2249/changes/21f8f20850125e60ec08d19503dbb2834e681582).

core/internal/initdb/initdb.go
gazerro: Restored in commit [Restore member name initialization without Docker](https://github.com/krenalis/krenalis/pull/2249/changes/11456e565b07633ba70ca3514b542cb40197a0f6).

Note that I also renamed a variable for consistency.

core/internal/initdb/upgrade_base58.go
gazerro: Fixed in commit [Align Base58 check constraint names with clean schema](https://github.com/krenalis/krenalis/pull/2249/changes/d886ecc48083cd9cb4741cc8a831ee984e48a211).

warehouses/snowflake/quote.go
gazerro: Changed in commit [Use strings.ContainsAny instead of IndexAny](https://github.com/krenalis/krenalis/pull/2249/changes/e4b00a2c2b37f31315fb842b27ad359cff497fb1).

core/internal/initdb/upgrade_base58.go
gazerro: Fixed in commit [Fix Base58 migration with events view](https://github.com/krenalis/krenalis/pull/2249/changes/ddec2fe906fa041ca7a5d25886496883a7117361).

zapateo (changes requested):
The various conversations are ok. 👍🏻 

Please, fix the conflict with `main`.

gazerro:
> The various conversations are ok. 👍🏻
> 
> Please, fix the conflict with `main`.

Done in commit [Merge branch 'main' into adopt-base58-ids](https://github.com/krenalis/krenalis/pull/2249/changes/8013fad36fd8a24579f359b59644293326ce25b1).

zapateo (approved):

PR #2254, opened 2026-05-29
core/internal: merge duplicated code in `destinations.go`

zapateo:
## Commit message

```
core/internal: merge duplicated code in `destinations.go`

Please note:

* When deleting an organization, a goroutine per workspace may be
  launched instead of a single goroutine, but this is acceptable.

* The "workspace has been deleted" message is passed to 'pipeline.Close'
  for both workspace and organization deletions. This was already the
  case and has been retained. Even if deleting an organization also
  implies deleting a workspace (so the message could still be correct as
  is), if we want to be more precise, we can change it in a separate
  commit.
```

### DIFF
--- a/core/internal/collector/destinations.go
+++ b/core/internal/collector/destinations.go
@@ -235,57 +235,14 @@ func (d *destinations) onDeletePipeline(n state.DeletePipeline) {
 
 // onDeleteOrganization is called when an organization is deleted.
 func (d *destinations) onDeleteOrganization(n state.DeleteOrganization) {
-	var pipelines []*destinationPipeline
 	for _, ws := range n.Organization().Workspaces() {
-		for _, c := range ws.Connections() {
-			if c.Role != state.Destination {
-				continue
-			}
-			connector := c.Connector()
-			if !connector.DestinationTargets.Contains(state.TargetEvent) {
-				continue
-			}
-			delete(d.senders, c.ID)
-			pipelines = append(pipelines, d.pipelines[c.ID]...) // No lock needed for reads while the state is frozen.
-			d.mu.Lock()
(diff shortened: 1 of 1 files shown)


core/internal/collector/destinations.go
```diff
@@ -394,6 +351,32 @@ func (d *destinations) onUpdatePipeline(n state.UpdatePipeline) {
 	d.mu.Unlock()
 }
 
+// removeWorkspace removes the workspace ws from destinations.
+func (d *destinations) removeWorkspace(ws *state.Workspace) {
```
gazerro, line 356: I'd add this comment because the other functions make this implicit in their names, but that's not obvious here.

```
// When this method is called, the state is frozen.
```

gazerro (changes requested):
> The "workspace has been deleted" message is passed to 'pipeline.Close'
  for both workspace and organization deletions. This was already the
  case and has been retained. Even if deleting an organization also
  implies deleting a workspace (so the message could still be correct as
  is), if we want to be more precise, we can change it in a separate
  commit.

A different message would be preferable, but I'll leave that up to you.

zapateo:
> A different message would be preferable, but I'll leave that up to you.

Okay, I'll mark it as a to-do and we'll do it in a separate commit.

core/internal/collector/destinations.go
zapateo: Done in [Add comment about frozen state](https://github.com/krenalis/krenalis/pull/2254/commits/71df481db1f6d2df6e8cbd47cb965c02a0a6209d).

I'll do the same change on the method added in the other PR, already merged.

gazerro (approved):

PR #2259, opened 2026-05-30
cmd: simplify `validateRequiredBody` using `http.MaxBytesReader`

gazerro:
```
cmd: simplify `validateRequiredBody` using `http.MaxBytesReader`

Previously, validateRequiredBody enforced the request body size limit
using io.LimitedReader and a custom reader type. Using
http.MaxBytesReader is more appropriate because it provides the
standard net/http behavior when the configured limit is exceeded.

Additionally, the previous implementation wrapped the request body with
io.NopCloser, causing Close to become a no-op instead of closing the
underlying request body.
```

### DIFF
--- a/cmd/apis_server.go
+++ b/cmd/apis_server.go
@@ -352,9 +352,7 @@ func (s *apisServer) forwardSentryError(w http.ResponseWriter, r *http.Request) 
 	if _, _, _, err := s.authenticateAdminRequest(r); err != nil {
 		return nil, err
 	}
-	lr := &io.LimitedReader{R: r.Body, N: maxRequestSize + 1}
-	payload := norm.NFC.Reader(lr)
-	r.Body = io.NopCloser(payload)
+	r.Body = http.MaxBytesReader(w, r.Body, maxRequestSize)
 	s.sentryTelemetry.errorTunnel.ServeHTTP(w, r)
 	return nil, nil
 }
--- a/cmd/apis_server.go
+++ b/cmd/apis_server.go
@@ -352,7 +352,9 @@ func (s *apisServer) forwardSentryError(w http.ResponseWriter, r *http.Request) 
 	if _, _, _, err := s.authenticateAdminRequest(r); err != nil {
 		return nil, err
 	}
-	r.Body = http.MaxBytesReader(w, r.Body, maxRequestSize)
+	lr := &io.LimitedReader{R: r.Body, N: maxRequestSize + 1}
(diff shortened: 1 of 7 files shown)


cmd/apis_server_test.go
```diff
@@ -402,19 +410,20 @@ func TestValidateRequiredBody(t *testing.T) {
...
 			t.Fatalf("expected nil error, got %v", err)
 		}
 		body, err := io.ReadAll(req.Body)
 		if err == nil {
 			t.Fatalf("expected error, got nil")
 		}
-		if err.Error() != "request body too large" {
-			t.Fatalf("expected request body too large, got %q", err.Error())
+		if _, ok := err.(*http.MaxBytesError); !ok {
+			t.Fatalf("expected http.MaxBytesError error, got %t", err)
```
zapateo: Maybe you meant `%T` instead of `%t`?

cmd/apis_server.go
```diff
@@ -352,16 +352,14 @@ func (s *apisServer) forwardSentryError(w http.ResponseWriter, r *http.Request)
 	if _, _, _, err := s.authenticateAdminRequest(r); err != nil {
 		return nil, err
 	}
-	lr := &io.LimitedReader{R: r.Body, N: maxRequestSize + 1}
-	payload := norm.NFC.Reader(lr)
```
zapateo, line 356: Removing normalization when forwarding requests to Sentry is not documented in the commit message, is this intentional?

zapateo (changes requested):

cmd/apis_server.go
zapateo: EDIT: I just saw the PR https://github.com/krenalis/krenalis/pull/2258, maybe its commit ended up in this branch by mistake?

cmd/apis_server_test.go
gazerro: Fixed in commit [Fix incorrect fmt verb in tests](https://github.com/krenalis/krenalis/pull/2259/changes/3eda501c810aeed62fc2237d37aa6323f8967b0e).

cmd/apis_server.go
gazerro: Fixed in commit [Revert unintended change](https://github.com/krenalis/krenalis/pull/2259/changes/1d5a1a475f3fb2e91aac5121c2b7d6c7010483b3).

zapateo (approved):

PR #2260, opened 2026-06-03
core: add transactional validation for disabled pipelines

gazerro:
```
core: add transactional validation for disabled pipelines

Previously, when starting a pipeline execution, the disabled-state check
was performed before the state-change transaction. As a result, a
concurrent start and disable operation could race, allowing a pipeline
to start while it was being disabled.

Add an additional disabled-state check within the transaction to prevent
a pipeline from being started while a disable operation is in progress.
```


core/pipeline.go
```diff
@@ -886,20 +888,23 @@ func (this *Pipeline) createRun(ctx context.Context, incremental *bool) (int, er
...
-			"LIMIT 1", n.Pipeline).Scan(&function, &inc, &cursor, &executing)
+			"LIMIT 1", n.Pipeline).Scan(&enabled, &function, &inc, &cursor, &executing)
 		if err != nil {
 			if err == sql.ErrNoRows {
 				return nil, errors.NotFound("pipeline %d does not exist", n.Pipeline)
 			}
 			return nil, err
 		}
+		if !enabled {
+			return 0, errors.Unprocessable(PipelineDisabled, "pipeline %d is disabled", this.pipeline.ID)
```
zapateo: In other cases we return `nil`, shouldn't we return it here too, instead of 0?

core/pipeline.go
zapateo, line 1: The behavior of a disabled pipeline is neither documented nor specified anywhere, from what I could find (correct me if I'm wrong).

Not even the API documentation, as far as I could find, mentions what disabling a pipeline entails: https://www.krenalis.com/docs/api/pipelines#set-status

So, beyond deducing from the code what is guaranteed for an enabled or disabled pipeline (and as we saw this morning, this isn't so trivial), it's difficult to understand what the APIs guarantee and what the core guarantees.

This is a discussion that goes beyond this PR, but I was wondering whether it was worth documenting at least a little. If it's too long and not worth the effort now, that's fine; we'll see about it later.

zapateo, line 1: This error message – although not touched by this PR – prints the connection ID instead of the pipeline ID if the pipeline is disabled:

https://github.com/krenalis/krenalis/blob/2d4369d99a6b126a4dd28d13d76560ee86a97513/core/pipeline.go#L516-L518

I think it's worth fixing on `main`.

zapateo (changes requested):

core/pipeline.go
gazerro: **Documentation:** a disabled pipeline cannot be run. This is documented in the API documentation at [https://www.krenalis.com/docs/api/pipelines#run-pipeline](https://www.krenalis.com/docs/api/pipelines#run-pipeline), where one of the possible error messages is `Pipeline is disabled`. In addition, on connector pages where the **Run** button is available, this is also mentioned. For example: *"Run the import immediately, one time only. Available only when the pipeline is enabled."* ([https://www.krenalis.com/docs/ingest-users/databases#pipelines](https://www.krenalis.com/docs/ingest-users/databases#pipelines)).

That said, I agree that this should also be stated explicitly in the documentation of the methods that change a pipeline's status.

**Specification:** strictly speaking, nothing in Krenalis is specified, since there is no formal specification. Did you mean something else by "specified"?

core/pipeline.go
gazerro: Good catch. Since the return value is expected to be a notification, returning `nil` is the correct behavior rather than returning `0`. Fixed in commit [Fix return value](https://github.com/krenalis/krenalis/pull/2260/changes/fd3e7584dbfca03d8e6135a0e71060832652aa3d).

core/pipeline.go
gazerro: Fixed in pull request https://github.com/krenalis/krenalis/pull/2261.

zapateo: Both regarding the documentation and the "specification" (meaning a brief comment in the code), consider what has been said verbally.

I'll leave this conversation open; you decide what to note, and you can close it whenever you like.

zapateo (approved):
PR approved.

Take a look at https://github.com/krenalis/krenalis/pull/2260/changes#r3354791012.

PR #2265, opened 2026-06-04
core/internal/collector: differentiate messages in `removeWorkspace`

zapateo:
## Commit message

```
core/internal/collector: differentiate messages in `removeWorkspace`
```

### DIFF
--- a/core/internal/collector/destinations.go
+++ b/core/internal/collector/destinations.go
@@ -7,6 +7,7 @@ package collector
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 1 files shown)


core/internal/collector/destinations.go
```diff
@@ -352,8 +353,11 @@ func (d *destinations) onUpdatePipeline(n state.UpdatePipeline) {
 }
 
 // removeWorkspace removes the workspace ws from destinations.
+// deletedResource names the resource whose deletion triggered the removal:
+// "workspace" when the workspace itself was deleted, or "organization" when an
+// organization was deleted (and with it all its workspaces).
 // When this method is called, the state is frozen.
-func (d *destinations) removeWorkspace(ws *state.Workspace) {
+func (d *destinations) removeWorkspace(ws *state.Workspace, deletedResource string) {
```
gazerro: This would be more readable if the cause were passed directly to the method:

```go
func (d *destinations) removeWorkspace(ws *state.Workspace, cause error)
```

and then used like this:

```go
pipeline.Close(cause)
```

This makes the dependency explicit and avoids relying on external state.

gazerro (changes requested):

core/internal/collector/destinations.go
zapateo: What do you think about something like this? [Pass the cause error instead of a string](https://github.com/krenalis/krenalis/pull/2265/commits/2b53afd147ac756566d68894401dbc009f0d877f)

The error is constructed inline and is not assigned to a variable for clarity of the code, the impact I believe is minimal (if not zero, as it is put inline by the compiler, I don't know), and also on the `main` it is constructed for each pipeline, so in any case it is an improvement from that point of view too.

gazerro: Even better, this makes it clear at the call site what the error represents:

```go
cause := errors.New("organization has been deleted")
d.removeWorkspace(ws, cause)
```

gazerro (changes requested):

core/internal/collector/destinations.go
zapateo: What about this? [Declare cause in variables to improve clarity](https://github.com/krenalis/krenalis/pull/2265/commits/38fcc03c04573a51cccb8aee159c5703755064e2)

gazerro (approved):

PR #2269, opened 2026-06-05
cmd: fail login when multiple organizations exist

gazerro:
```
cmd: fail login when multiple organizations exist

The login flow assumes that exactly one organization exists.

Instead of looking up the member in the first organization sorted
alphabetically, fail with an internal error when multiple organizations
are present.
```

### DIFF
--- a/cmd/apis_server.go
+++ b/cmd/apis_server.go
@@ -376,7 +376,7 @@ func (s *apisServer) login(w http.ResponseWriter, r *http.Request) (any, error) 
 		return nil, errors.New("there are no organizations")
 	}
 	if len(organizations) > 1 {
-		return nil, errors.New("there are more than one organizations")
+		return nil, errors.New("there is more than one organization")
 	}
 	org := organizations[0]
 	memberID, err := org.AuthenticateMember(r.Context(), body.Email, body.Password)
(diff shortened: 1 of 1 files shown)


cmd/apis_server.go
```diff
@@ -375,6 +375,9 @@ func (s *apisServer) login(w http.ResponseWriter, r *http.Request) (any, error)
 	if len(organizations) == 0 {
 		return nil, errors.New("there are no organizations")
 	}
+	if len(organizations) > 1 {
+		return nil, errors.New("there are more than one organizations")
```
zapateo: I'm not sure if this is correct in English. Shouldn't it be like:

> there is more than one organization

?

cmd/apis_server.go
zapateo: A `limit` of 1 is passed here, so the number of organizations returned is at most 1, and the subsequent check (introduced by this PR)  always fails. Or am I missing something?

I also noticed we have the `Core.CountOrganizations` method. If so, we can evaluate whether it makes sense to use that.

zapateo (changes requested):

cmd/apis_server.go
gazerro: Fixed in commit [Fix error message](https://github.com/krenalis/krenalis/pull/2269/changes/6ae6e063886beb9d827bb88f5d6e5defa9e12eac).

cmd/apis_server.go
gazerro: Fixed in commit [Fix length check](https://github.com/krenalis/krenalis/pull/2269/changes/f6bf7aa67afef0b27514b1f02ea338d68c559464).

zapateo (approved):

PR #2275, opened 2026-06-10
core: prevent concurrent runs for the same pipeline

gazerro:
```
core: prevent concurrent runs for the same pipeline

Ensure that each pipeline can have at most one active run.

The existing check was insufficient because concurrent transactions
could both observe no active run before inserting their respective runs.
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -38,6 +38,7 @@ func Main(assets fs.FS) {
 	var configStore string
 	var initDBIfEmpty bool
 	var initDockerMember bool
+	var upgradeDB bool
 	flag.BoolVar(&help, "help", false, "print the help for krenalis and exit")
 	flag.StringVar(&configStore, "config-store", "env:",
 		"configuration source: 'env:' to read KRENALIS_* from the environment, or 'aws:<region>:<prefix>' to read from AWS Parameter Store (default: 'env:')")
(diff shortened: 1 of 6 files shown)


zapateo (changes requested):
Let's see if I'm understanding the issue we're having on the `main` branch. Correct me if I'm wrong.

I'm seeing that `Pipeline.createRun`, which checks if a pipeline is already running, opens a transaction that first does a SELECT and then an INSERT.

The transaction has a "READ COMMITTED" isolation level, which is the default. This is also documented here:

https://github.com/krenalis/krenalis/blob/6c9d1fc7330d663dc051d8dd55c559d166deaa59/core/internal/db/db.go#L135-L144

So, if I'm correctly interpreting the [first paragraph of the PostgreSQL Read Committed Isolation Level documentation](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED), this isolation level doesn't guarantee that if a SELECT statement detects that a pipeline is running (and therefore an INSERT can be made to start it), another transaction can't also verify that the pipeline isn't running and then perform an INSERT and commit the change.

As a result, two lines with different IDs can be inserted, effectively executing the same pipeline twice.

Is what I said correct?

gazerro:
> Is what I said correct?

Yes, it's correct.

gazerro:
As discussed, the transaction handling code has been simplified in commit [Simplify transaction code](https://github.com/krenalis/krenalis/pull/2275/changes/55f081a06392d952f4a78e0550e7ccb44a035d30).

zapateo (approved):

PR #2278, opened 2026-06-15
core: fix active pipeline run detection after Base58 migration

gazerro:
```
core: fix active pipeline run detection after Base58 migration

Commit 6c9d1fc73 introduced random Base58 identifiers for pipeline runs,
but createRun still assumed that the highest ID corresponded to the most
recent run. As a result, it could select a completed run and miss an
active one. Detect active runs by checking for a NULL end time instead
of relying on ID ordering.

While at it, rename the pipelines_runs alias from "e" to "r" for
clarity.
```


gazerro:
Following the changes introduced by commit 55f081a in pull request #2275, this pull request has become redundant and can be safely closed.

PR #2281, opened 2026-06-16
core: fix scheduler executor restart after leadership loss

gazerro:
```
core: fix scheduler executor restart after leadership loss

Previously, the pipeline scheduler kept a reference to its executor
after a node lost leadership, so the same node could not create a new
executor if it became leader again.

This change clears the executor reference when leadership is lost. Since
the executor can now be nil after leadership changes, it also captures
the current executor before starting asynchronous removal operations.
```

### DIFF
--- a/core/pipeline_scheduler.go
+++ b/core/pipeline_scheduler.go
@@ -96,11 +96,11 @@ func (ps *pipelineScheduler) onDeleteConnection(n state.DeleteConnection) {
 	if pipelines == nil {
 		return
 	}
-	go func() {
+	go func(e *pipelineExecutor) {
 		for _, pipeline := range pipelines {
-			ps.executor.RemovePipeline(pipeline)
+			e.RemovePipeline(pipeline)
 		}
-	}()
+	}(ps.executor)
 }
 
 // onDeleteOrganization is called when an organization is deleted from the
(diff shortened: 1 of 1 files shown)


core/pipeline_scheduler.go
zapateo, line 1: The issue seems broader than what is indicated in the commit message, so I would extend that message to make it more general.

core/pipeline_scheduler.go
```diff
@@ -96,11 +96,11 @@ func (ps *pipelineScheduler) onDeleteConnection(n state.DeleteConnection) {
 	if pipelines == nil {
 		return
 	}
-	go func() {
```
zapateo, line 99: These changes, which immediately evaluate `ps.executor`, fix a sort of "latent bug" that never occurred before due to the other bug in `onElectLeader`.

I'm not sure if it's worth mentioning in the commit message; you decide what you think is clearer.

zapateo (changes requested):

core/pipeline_scheduler.go
gazerro: Isn't it already?

>  Since the executor can now be nil after leadership changes, it also captures the current executor before starting asynchronous removal operations.

core/pipeline_scheduler.go
gazerro: I see your point, but I would keep the commit message focused on the issue fixed here.

The old executor reference was kept after leadership loss, and its internal state could still be updated. However, once the executor is closed, no pipeline execution happens through it anymore, so this does not seem to cause any visible unwanted behavior.

The actual bug is that keeping that reference prevents the same node from creating a new executor if it becomes leader again. Since there is no code change specifically for the internal state updates of the closed executor, I would avoid mentioning that part in the commit message.

core/pipeline_scheduler.go
zapateo: Of course, you're right. I hadn't noticed it at first, and then I assumed it wasn't there.

zapateo (approved):

PR #2283, opened 2026-06-17
core: fix duplicate member email error handling

gazerro:
```
core: fix duplicate member email error handling

Previously AddMember checked for duplicate emails before inserting a
member, leaving a race where the insert could still fail with a generic
unique constraint error.

Now it relies on the database constraint and maps duplicate email
violations to MemberEmailExists.
```

### DIFF
--- a/core/organization.go
+++ b/core/organization.go
@@ -185,13 +185,6 @@ func (this *Organization) AddMember(ctx context.Context, member MemberToSet) err
 		})
 		now := time.Now().UTC()
 		err = this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
-			exists, err := tx.QueryExists(ctx, "SELECT FROM members WHERE organization = $1 AND email = $2", this.organization.ID, member.Email)
-			if err != nil {
-				return nil, err
-			}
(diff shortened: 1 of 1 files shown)


core/organization.go
```diff
@@ -185,13 +185,6 @@ func (this *Organization) AddMember(ctx context.Context, member MemberToSet) err
 		})
 		now := time.Now().UTC()
 		err = this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
```
zapateo, line 187: Is there any benefit to continuing to use a transaction here? Can't it just be removed now?

zapateo (changes requested):

core/organization.go
gazerro: Yes. Note that this is using `this.core.state.Transaction`, not `this.core.db.Transaction`. That's intentional because, after the queries shown here, we also create the notification in the database and send it through `notify`. Those steps should stay in the same state transaction as the queries above.

zapateo: Ok, clear!

zapateo (approved):

PR #2286, opened 2026-06-19
core: make scheduler leader transitions explicit

gazerro:
```
core: make scheduler leader transitions explicit

This makes the executor creation and shutdown paths easier to follow
without changing behavior.
```


zapateo (changes requested):
This PR has a conflict with `main`. Could you please resolve it? Then I can proceed with the review.

zapateo (approved):
Much more readable now.

PR #2292, opened 2026-06-23
core: fix pipeline run metrics

gazerro:
```
core: fix pipeline run metrics

Previously, when a pipeline run started, the metrics for the run start
timeslot were saved as a baseline to exclude metrics that already
existed before the run.

When the run ended, metrics from all timeslots starting from that
timeslot were summed and adjusted by subtracting the saved baseline.

This worked only while those timeslots remained unaggregated. Since
timeslots older than 60 minutes are aggregated into earlier buckets,
runs lasting more than 60 minutes could end up with incorrect metric
counts.

For example, if a run started in timeslot 125 and wrote metrics in
timeslot 126, those metrics could later be aggregated into timeslot 120.
When the run ended, only metrics from timeslot 125 onward were counted,
so the metrics moved to timeslot 120 were skipped.

Fix this by saving the total pipeline metrics at run start, instead of
only the metrics for the start timeslot. This makes the final run
metrics independent of timeslot aggregation.
```

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -1241,7 +1241,6 @@ func (core *Core) tryStartPipelineRun(pipelineID string) {
 		defer stopPing()
 
 		// Prepare the run metrics.
-		timeSlot := metrics.TimeSlotFromTime(run.StartTime)
 		bo = backoff.New(200)
 		for bo.Next(ctx) {
 			_, err := core.db.Exec(ctx,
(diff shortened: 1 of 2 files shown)


zapateo (approved):
PR approved. Regarding the error handling we discussed verbally, I'll make a separate change after this PR is merged.

PR #2296, opened 2026-06-23
core: handle missing organization when adding members

gazerro:
```
core: handle missing organization when adding members

Return OrganizationNotExist when inserting a member fails because the
organization no longer exists.

While I was there, keep the INSERT column order aligned with the members
table schema.
```

### DIFF
--- a/core/organization.go
+++ b/core/organization.go
@@ -163,9 +163,11 @@ type AccessKey struct {
 
 // AddMember adds a new member of the organization.
 //
-// If the member to add has an email that is already used by another
-// member, it returns an errors.UnprocessableError error with code
-// MemberEmailExists.
+// It returns an errors.UnprocessableError error with code
+//
+//   - MemberEmailExists, if a member with this email already exists in the
+//     organization.
+//   - OrganizationNotExist, if the organization does not exist.
 func (this *Organization) AddMember(ctx context.Context, member MemberToSet) error {
 	this.core.mustBeOpen()
 	err := validateMemberToSet(member, true, true, true)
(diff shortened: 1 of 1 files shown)


core/organization.go
```diff
@@ -187,14 +189,17 @@ func (this *Organization) AddMember(ctx context.Context, member MemberToSet) err
...
 				_, err = tx.Exec(ctx,
-					"INSERT INTO members (id, name, email, password, avatar.image, avatar.mime_type, organization, created_at) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)",
-					n.ID, member.Name, member.Email, password, member.Avatar.Image, member.Avatar.MimeType, this.organization.ID, now)
+					"INSERT INTO members (id, organization, name, avatar.image, avatar.mime_type, email, password, created_at) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)",
+					n.ID, this.organization.ID, member.Name, member.Avatar.Image, member.Avatar.MimeType, member.Email, password, now)
 			} else {
 				_, err = tx.Exec(ctx,
-					"INSERT INTO members (id, name, email, password, avatar, organization, created_at) VALUES ($1, $2, $3, $4, $5, $6, $7)",
-					n.ID, member.Name, member.Email, password, nil, this.organization.ID, now)
+					"INSERT INTO members (id, organization, name,  avatar, email, password, created_at) VALUES ($1, $2, $3, $4, $5, $6, $7)",
```
zapateo: There is a double space here, between `name,` and `avatar`.

zapateo (changes requested):

core/organization.go
gazerro: Fixed in commit [Fix double space](https://github.com/krenalis/krenalis/pull/2296/changes/7caefee7512e070e607112d9178c2de1ff8f90e5).

zapateo (approved):

PR #2306, opened 2026-06-25
core: avoid starting the wrong pipeline run

gazerro:
```
core: avoid starting the wrong pipeline run

Previously, a delayed pipeline startup goroutine could start the wrong 
run if a pipeline completed and another run for the same pipeline
started before the goroutine looked up the current pipeline run.

Start pipeline runs using the active run, now called the live run,
selected by the notification or startup scan, so the goroutine keeps
targeting the same run. To support this, track live runs in shared state
by run ID and rename the run completion helper to match that flow.
```

### DIFF
--- a/core/internal/state/keep.go
+++ b/core/internal/state/keep.go
@@ -619,7 +619,6 @@ func (state *State) createWorkspace(n notification) string {
 	ws := Workspace{
 		mu:                             &sync.Mutex{},
 		connections:                    map[string]*Connection{},
-		runs:                           map[string]*PipelineRun{},
 		ID:                             e.ID,
 		organization:                   organization,
 		Name:                           e.Name,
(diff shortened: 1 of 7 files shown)


core/internal/state/keep.go
zapateo, line 1: We use the term "active run" here:

https://github.com/krenalis/krenalis/blob/3b5219fd015dc8ec7a3cc775b1fcf99641562934/core/internal/initdb/schema.sql#L190-L192

It should be renamed to be coherent with the term "live run" introduced by this PR.

Logically, it makes sense to do it in this PR, the only thing is that it's a change that also affects the updater, so I think it's cleaner to do it all in a separate commit on `main`. Do what you think is best.

zapateo, line 1: The "run in progress" expression, which we expose externally:

https://github.com/krenalis/krenalis/blob/803920af6d6ebba381194f22e5c3b47d74de815b/core/pipeline.go#L495-L497

Conflicts with the concept of "live run". We can use "live run" only internally and rename the `RunInProgress` error (as discussed) to something like "pipeline already running". 

Note that the documentation on krenalis.com also needs to be updated. Decide whether to make the change in this PR or in a separate one, It's fine with me in any case.

zapateo, line 1: Remember to update the commit message, introducing the term "live run" before using it.

zapateo (changes requested):

core/internal/state/keep.go
gazerro: See pull request [core: rename active pipeline run index to live](https://github.com/krenalis/krenalis/pull/2308).

gazerro: Done.

gazerro: See pull requests:
- [core: rename pipeline in-progress error to `PipelineAlreadyRunning`](https://github.com/krenalis/krenalis/pull/2309)
- [Update pipeline already-running API error code](https://github.com/krenalis/krenalis.com/pull/38).

gazerro:
See also commit [Fix comment about loading live pipeline runs](https://github.com/krenalis/krenalis/pull/2306/changes/b3e7bf74d639dc71ab8ec1a334b14481835014df).

zapateo (approved):

PR #2308, opened 2026-06-25
core: rename active pipeline run index to live

gazerro:
```
core: rename active pipeline run index to live

Rename the partial unique index that prevents concurrent pipeline runs
from pipelines_one_active_run_idx to pipelines_one_live_run_idx.
```

### DIFF
--- a/core/internal/initdb/upgrade.go
+++ b/core/internal/initdb/upgrade.go
@@ -41,7 +41,7 @@ func Upgrade(ctx context.Context, database *db.DB) error {
 		if _, err := tx.Exec(ctx, `DROP INDEX IF EXISTS `+oneActivePipelineRunIndex); err != nil {
 			return err
 		}
-		// core,core/internal/inidb: Rename active pipeline run index to live
+		// core: rename active pipeline run index to live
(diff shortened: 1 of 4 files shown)


core/internal/initdb/upgrade.go
```diff
@@ -38,11 +41,16 @@ func Upgrade(ctx context.Context, database *db.DB) error {
 		if _, err := tx.Exec(ctx, `DROP INDEX IF EXISTS `+oneActivePipelineRunIndex); err != nil {
 			return err
 		}
-		if _, err = tx.Exec(ctx, `CREATE UNIQUE INDEX `+oneActivePipelineRunIndex+`
+		// core,core/internal/inidb: Rename active pipeline run index to live
```
zapateo: Just a minor typo here, `inidb` instead of `initdb`.

This is not so relevant to me, considering that this code will be discarded. Feel free to change it or leave it as it is.

zapateo (dismissed):

core/internal/initdb/upgrade.go
gazerro: Done.

PR #2310, opened 2026-06-26
core: cancel local pipeline runs when the cleaner ends them

gazerro:
```
core: cancel local pipeline runs when the cleaner ends them

Previously, the pipeline cleaner could mark an orphaned run as ended
while the local goroutine kept executing the import or export until
completion.

Store a cancel function for each pipeline run started by this node.
When a run ends, cancel the matching local context so the import or
export can stop.

This is still cooperative cancellation: preventing a new run from
starting until the previous goroutine has fully exited is left for a
follow-up change.

For #2250
```

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -64,6 +64,7 @@ type Core struct {
 	functionProvider  transformers.FunctionProvider
 	pipelineCleaner   *pipelineCleaner
 	pipelineScheduler *pipelineScheduler
+	localLiveRuns     cancelContexts
 	memberEmailFrom   string
 	smtp              *SMTPConfig
 	close             struct {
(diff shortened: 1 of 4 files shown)


zapateo (changes requested):
This PR has conflicts with `main`, could you please resolve them?

gazerro:
> This PR has conflicts with main, could you please resolve them?

Resolved.

core/core.go
```diff
@@ -1194,13 +1196,106 @@ func (core *Core) encryptAuthorizedOAuthAccount(ctx context.Context, account aut
...
+					// The pipeline does not exist anymore.
+					return nil, nil
+				}
+				return nil, err
+			}
+			return n, nil
+		})
+		if err != nil {
+			if ctx.Err() == nil {
+				slog.Error("core: cannot end pipeline run; retrying", "retry_after", bo.WaitTime(), "error", reason)
```
zapateo: Is it correct to log `reason` here? Shouldn't it be `err` instead?

zapateo (changes requested):

core/core.go
gazerro: Fixed in commit [Fix logged message](https://github.com/krenalis/krenalis/pull/2310/changes/e6f3b1648496273d38d2741f49bc844f362af2e9).

zapateo (approved):

PR #2311, opened 2026-06-26
Handle disabled organizations in the Admin

zapateo:
Please **review these commits individually**, as they will be pushed back to `main` one by one, separately (via _Rebase and merge_).

### DIFF
--- a/admin/src/components/routes/App/App.tsx
+++ b/admin/src/components/routes/App/App.tsx
@@ -13,7 +13,7 @@ import type SlAlert from '@shoelace-style/shoelace/dist/components/alert/alert.j
 import SlSpinner from '@shoelace-style/shoelace/dist/react/spinner/index.js';
 import '@shoelace-style/shoelace/dist/themes/light.css';
 import { useApp } from './useApp';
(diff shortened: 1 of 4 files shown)


retini (commented):
@zapateo, regarding the description of commit 367cbf5c502d8fd23e237f3b8fd5b73c0a445f9f:

> passwordless login does not work silently, and it's unclear why.

Note that the reason why `tryPasswordlessLogin` currently doesn't show any errors is that we have no way of knowing in advance whether a given Krenalis version supports passwordless login. For this reason, we attempt the requests and intentionally suppress any errors if they fail, since the failure may simply indicate that the target Krenalis instance does not support passwordless authentication.

That said, it is correct to show the `OrganizationDisabled` error because it is returned by the server only after the passwordless authentication has completed successfully, so we know for sure that the Krenalis instance supports it.

The only change I would suggest is updating the commit description and optionally adding a comment explaining why the other errors are intentionally suppressed, if you think that would make the logic clearer for future readers.

Other than that, the PR looks good to me and I approve the change.

PR #2316, opened 2026-06-27
connectors/dummy: replace simulated HTTP delay with `operationDelay`

gazerro:
```
connectors/dummy: replace simulated HTTP delay with `operationDelay`

Previously, Dummy exposed a simulated HTTP delay setting.

Replace it with operationDelay, a Go-style duration string such as
"500ms", "2s", or "1m". This lets users choose an explicit delay for
Dummy operations, including real event dispatches, instead of only
enabling a fixed simulated delay.

operationDelay also accepts ranges such as "2s-5m", meaning each
operation uses a random delay between 2 seconds and 5 minutes.

A missing or empty value means no delay, and the maximum accepted delay
is "24h". Existing Dummy connections with simulated HTTP delay enabled
are migrated to operationDelay: "2s".
```

### DIFF
--- a/connectors/dummy/dummy.go
+++ b/connectors/dummy/dummy.go
@@ -340,7 +340,7 @@ func (dummy *Dummy) ServeUI(ctx context.Context, event string, settings json.Val
 				Name:        "operationDelay",
 				Label:       "Operation delay",
 				Placeholder: "500ms",
-				HelpText:    "If set, Dummy adds the specified delay to each operation. Examples: 500ms, 2s, 1m, 2s-5m.",
+				HelpText:    "Delay added to each operation. Use a fixed duration, such as 500ms, 2s, or 1m, or a range for a random delay, such as 2s-5m or 30m-2h.",
 				Role:        connectors.Both,
 			},
 		},
--- a/connectors/dummy/dummy.go
+++ b/connectors/dummy/dummy.go
@@ -568,7 +568,7 @@ func parseOperationDelay(operationDelay string) (string, error) {
 			return "", nil
 		}
 		if minDelay > maxOperationDelay {
-			return "", connectors.NewInvalidSettingsError("operation delay must be less than or equal to 24h")
+			return "", connectors.NewInvalidSettingsError(fmt.Sprintf("maximum operation delay must be less than or equal to %s", maxOperationDelay))
 		}
 		return minDelay.String(), nil
 	}
@@ -586,7 +586,7 @@ func parseOperationDelay(operationDelay string) (string, error) {
 		return "", nil
 	}
 	if maxDelay > maxOperationDelay {
-		return "", connectors.NewInvalidSettingsError("maximum operation delay must be less than or equal to 24h")
+		return "", connectors.NewInvalidSettingsError(fmt.Sprintf("maximum operation delay must be less than or equal to %s", maxOperationDelay))
 	}
 	if minDelay > maxDelay {
 		return "", connectors.NewInvalidSettingsError("minimum operational delay cannot be greater than maximum delay")
--- a/connectors/dummy/dummy.go
+++ b/connectors/dummy/dummy.go
@@ -485,11 +485,10 @@ func (dummy *Dummy) saveSettings(ctx context.Context, settings json.Value) error
 	}
 	// Validate operation delay.
 	if s.OperationDelay != "" {
-		delay, err := parseOperationDelay(s.OperationDelay)
+		err = parseOperationDelay(s.OperationDelay)
 		if err != nil {
 			return err
 		}
-		s.OperationDelay = delay
 	}
 	return dummy.env.Settings.Store(ctx, s)
 }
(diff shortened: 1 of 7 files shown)


core/internal/initdb/upgrade.go
zapateo, line 1: I don't know if it's worth introducing the updater code, which is additional code anyway, perhaps causing problems during the update, etc., and I don't know what advantage it provides.

In fact, we don't have a Dummy connector in our Krenalis instance, and for local installations, cleaning the installation is probably fine. What do you think? If you'd like, we can discuss it face-to-face.

connectors/dummy/dummy.go
```diff
@@ -416,6 +435,23 @@ func (dummy *Dummy) Upsert(ctx context.Context, target connectors.Targets, recor
 	return nil
 }
 
+// applyOperationDelay applies an operation delay.
+func (dummy *Dummy) applyOperationDelay(ctx context.Context, operationDelay string) error {
```
zapateo, line 440: The timeout introduced now is constant, unlike the version we have in `main` which simulates various and random timeouts, which creates random test scenarios.

I'd like to better understand what the advantage of this change is.

zapateo (changes requested):

core/internal/initdb/upgrade.go
gazerro: After thinking about it, I'm probably a bit more conservative than you on this point, but I think the main reason for removing it is that Dummy is only used for development, so it shouldn't be part of database updates.

As discussed, I've removed the database update in this commit: [Remove database update](https://github.com/krenalis/krenalis/pull/2316/changes/0f2e39332a1d1648129d48079c69639e63dd6e6d).

connectors/dummy/dummy.go
gazerro: > The timeout introduced now is constant, unlike the version we have in main which simulates various and random timeouts, which creates random test scenarios.

As discussed, I added support for duration ranges, so the delay can be randomized within the configured interval: [Support ranged operation delays](https://github.com/krenalis/krenalis/pull/2316/changes/83a675fb4068c4c3b9c5e89cd530179bb583c8ac).

> I'd like to better understand what the advantage of this change is.

The main purpose of this change is to let us verify that disabling an organization stops a batch pipeline from continuing to run.

connectors/dummy/dummy.go
```diff
@@ -507,3 +552,47 @@ func deepClone(properties map[string]any) map[string]any {
...
+
+// parseOperationDelay parses and canonicalizes an operation delay.
+func parseOperationDelay(operationDelay string) (string, error) {
+	minPart, maxPart, hasMax := strings.Cut(operationDelay, "-")
+	if minPart == "" {
+		return "", connectors.NewInvalidSettingsError("operation delay must be a valid duration")
+	}
+	minDelay, err := time.ParseDuration(minPart)
+	if err != nil {
+		return "", connectors.NewInvalidSettingsError("operation delay must be a valid duration")
```
zapateo: In case I set an operation delay like:

```
2s-2min
```

(note that I mistakenly entered _min_ instead of _m_) the error displayed in the Admin is:

```
operation delay must be a valid duration
```

when in fact the operation delay can be either (1) a duration, like the error states or (2) a range of two durations (that's what I meant).

I would have expected something like:

```
operation delay must be a valid duration or interval
```

or something like that.

connectors/dummy/dummy.go
```diff
@@ -321,10 +336,12 @@ func (dummy *Dummy) ServeUI(ctx context.Context, event string, settings json.Val
...
 			},
-			&connectors.Checkbox{
-				Name:  "simulateHTTPDelay",
-				Label: "Pretend that Dummy operates via HTTP calls, introducing fictitious delays",
-				Role:  connectors.Both,
+			&connectors.Input{
+				Name:        "operationDelay",
+				Label:       "Operation delay",
+				Placeholder: "500ms",
+				HelpText:    "If set, Dummy adds the specified delay to each operation. Examples: 500ms, 2s, 1m, 2s-5m.",
```
zapateo: In the Admin interface, in the Dummy settings, the only reference to the operation delay being an interval is in the last example and the fact that it contains a hyphen. This tends to get lost:

immagine

If you can think of a way, I would make it a little more explicit, I don't know, maybe adding something to the help text of the text input.

connectors/dummy/dummy.go
zapateo: Got it. Please, remember to update the commit message.

connectors/dummy/dummy.go
```diff
@@ -85,6 +86,8 @@ var (
 //go:embed customers.json
 var jsonCustomers []byte
 
+const maxOperationDelay = 24 * time.Hour // maximum operation delay
```
zapateo, line 89: A constant is introduced here, which is then used in the comparisons in `parseOperationDelay`. However, the error messages don't use the constant, but "24h" is encoded directly in the error messages.

I'm concerned that introducing a constant that is used in some places but not others could create inconsistencies, as perhaps the constant is changed (giving the impression that everything has been changed accordingly) but the old error messages remain. In that case, perhaps I'd prefer not to have a constant at all, and so I'm forced to search for references to "24 hours" throughout the connector.

Anyway, I think we can leave it as is; Dummy is just a test connector, so it's not critical. Do what you prefer.

zapateo (changes requested):

connectors/dummy/dummy.go
gazerro: The commit [Improve error messages](https://github.com/krenalis/krenalis/pull/2316/changes/2149c781cfd6ac85072de811bae1ca1934405244) makes the error messages more precise. For `2s-2min`, the error returned now is

```
maximum operation delay must be a valid duration
```

connectors/dummy/dummy.go
gazerro: Fixed in commit [Improve UI help text](https://github.com/krenalis/krenalis/pull/2316/changes/fc1dc4f17bf74df66cd95712fc3126867e2b1e3c).

connectors/dummy/dummy.go
gazerro: Good point. I preferred to keep the constant and update the error messages to use it: [Derive max delay validation message from constant](https://github.com/krenalis/krenalis/pull/2316/changes/6c919a7a6563470b248299fde1677bf9b056117f).

zapateo (changes requested):
If I enter something like this in the Dummy settings in the Admin:

```
2m-5m
```

Then I save and reload the page, the value becomes:

```
2m0s-5m0s
```

That's because the **String** method of **Duration** is called.

As discussed verbally, we can avoid calling String and leave the value as entered by the user.

gazerro:
> As discussed verbally, we can avoid calling String and leave the value as entered by the user.

Done in commit [Preserve delay string format](https://github.com/krenalis/krenalis/pull/2316/changes/b8519ae53b8ae3935e2abdcf60d2b16f12e7abb0).

gazerro:
See also commit [Use NewInvalidSettingsErrorf when appropriate](https://github.com/krenalis/krenalis/pull/2316/changes/01bfc441539e83b473de054f47878b91241e56bb).

zapateo (approved):

PR #2317, opened 2026-06-27
core: cancel live runs when owning resources are removed or disabled

### LINKED ISSUE #2250: Stop batch pipeline execution when a pipeline is disabled or deleted
Currently, **when a pipeline is disabled or deleted, the pipeline's execution isn't interrupted**, so the behavior isn't well managed (does it continue to the end? Does it fail if the store is deleted?).

We should therefore:

1. **Implement a mechanism** to ensure that a pipeline's execution can be interrupted.
2. **Determine the circumstances** that should cause a pipeline to be interrupted.

Regarding point 2, the following circumstances can be identified:

* a pipeline is disabled.
* a pipeline is deleted (this includes deleting connections, workspaces, and organizations).
* an organization is disabled (wait for https://github.com/krenalis/krenalis/pull/2248).

This needs to be further explored and implemented.

Regarding the organization disabling case, this issue is crucial to ensuring that a pipeline's execution is effectively blocked, so the issue is critical. (wait for https://github.com/krenalis/krenalis/pull/2248)

gazerro:
```
core: cancel live runs for deleted owners and disabled organizations

Previously, live pipeline runs could keep executing after their owning
pipeline, connection, workspace, or organization was removed from state.
Disabling an organization also prevented new work from starting, but did
not explicitly stop runs that were already live.

Cancel live runs when their owning resources are deleted. When an
organization is disabled, close its live runs in the database as part of
the same transaction. Then update state from the notification and cancel
the matching local executions so connectors can stop promptly.

Closes #2250
```

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -389,11 +389,14 @@ func New(ctx context.Context, conf *Config) (_ *Core, err error) {
 	// Listen to state changes.
 	core.state.Freeze()
 	core.state.AddListener(core.onCreateWorkspace)
+	core.state.AddListener(core.onDeleteConnection)
 	core.state.AddListener(core.onDeleteOrganization)
+	core.state.AddListener(core.onDeletePipeline)
 	core.state.AddListener(core.onDeleteWorkspace)
 	core.state.AddListener(core.onElectLeader)
 	core.state.AddListener(core.onEndPipelineRun)
 	core.state.AddListener(core.onRunPipeline)
+	core.state.AddListener(core.onSetOrganizationStatus)
 	core.state.AddListener(core.onStartAlterProfileSchema)
 	core.state.AddListener(core.onStartIdentityResolution)
 	core.state.AddListener(core.onUpdateWarehouse)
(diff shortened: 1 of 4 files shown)


core/core.go
zapateo, line 1: _(I'm adding the note here because I don't know where else to add it.)_

In the commit message, we should replace:

```
cancellable connectors
```

with just:

```
connectors
```

test/organization_disabled_test.go
```diff
@@ -102,9 +102,66 @@ func TestOrganizationDisabled(t *testing.T) {
 	// Create an API key while the organization is still enabled.
 	apiKey := k.CreateWorkspaceRestrictedAPIKey("Events ingestion key")
 
+	longRunningPipelineParams := func(name string) krenalistester.PipelineToSet {
```
zapateo, line 105: This test covers the case of a pipeline cancelling when the organization is disabled.

How much would it cost to add a separate test, testing for example a the cancelling in case of a pipeline and/or connection deletion? I don't know if and how complex such a test would be to implement in practice, but I'm throwing out the idea because it might be useful. 

Evaluate whether it's worth it or not.

zapateo (changes requested):

core/core.go
gazerro: Fixed.

test/organization_disabled_test.go
gazerro: That makes sense as a thing to consider. The main complication is that, before this pull request, a batch run could still stop, but only once it reached a point where it noticed that a required resource, for example the run in the database, was no longer present.

This change instead cancels the run context, so the run can stop as soon as the code checks that context. In general, this should happen earlier than the previous case, but that is not strictly guaranteed.

Because of that, distinguishing in a test whether the run stopped because the context was actually cancelled, rather than because some resource was no longer present, is quite tricky and may not always be feasible. Also, even if the run stopped because a resource was missing, that would not necessarily prove that the context cancellation did not happen.

So I agree the test would be useful in principle, but at the moment I do not see a simple and reliable way to test this case separately.

zapateo (approved):

PR #2319, opened 2026-06-27
krenalis: update `github.com/ClickHouse/clickhouse-go` to v2.47.0

gazerro:
I asked Codex to inspect the vendor code:

> Inspect the vendored dependency changes caused by the github.com/ClickHouse/clickhouse-go/v2  update and flag any suspicious, malicious, or unexpected code.                                           

```
I inspected the staged vendored dependency update and did not find evidence of malicious code: no new runtime process execution, covert network calls, init hooks, credential/env harvesting, or suspicious file writes. go mod verify passed.

  Review-worthy items:

  - vendor/github.com/ClickHouse/clickhouse-go/v2/clickhouse_options.go:453 now sends debug logs to stdout when Debug=true and no custom logger/Debugf is set, via vendor/github.com/ClickHouse/clickhouse-go/v2/logger.go:112. Not malicious, but it can unexpectedly emit query/connection details to stdout in
    production if DSN debug=1 is used.

  - vendor/github.com/ClickHouse/clickhouse-go/v2/AI_POLICY.md:5 is a new upstream policy doc saying AI use need not be disclosed. Not executable, but unusual enough to note in a supply-chain review.
  - vendor/github.com/ClickHouse/clickhouse-go/v2/conn.go:324 now marks the connection closed for all flush errors, not just broken-pipe/EOF cases. This looks like an intended resiliency fix, not suspicious.
  - vendor/github.com/ClickHouse/clickhouse-go/v2/clickhouse.go:126 removes the embedded contributors list and returns an empty slice. Expected cleanup, but technically a behavior change.

  The broader dependency fan-out is expected from clickhouse-go/v2 v2.46.0 -> v2.47.0: ch-go, Docker/Moby client pieces, OpenTelemetry, compression libs, x/net, and a few platform helper libs moved too. The only check failure was git diff --cached --check, caused by trailing whitespace in vendored markdown
  files, mostly upstream docs/changelogs.
```


zapateo (approved):

zapateo (changes requested):
I initially approved this PR, but while working on something else, I realized there's a problem with the vendor and with Git, perhaps related to some modified files. I'll investigate a bit before approving it again.

zapateo (approved):
After further investigation, it was discovered that the issue is not with this PR, but with the Gitattributes configuration, which will be handled separately. PR approved.

PR #2323, opened 2026-06-29
cmd, core: fix not initialized database error handling

zapateo:
## Commit message

```
cmd, core: fix not initialized database error handling

Previously, when a Krenalis database hadn't been initialized, a specific
error was printed that explained how to resolve the issue.

This no longer works; the commit that broke this behavior is the
following: https://github.com/krenalis/krenalis/commit/c677d9408cf61ca02ea53b7704f0d81350c4daf3

The problem is that the captured error is now embedded into a string
instead:

https://github.com/krenalis/krenalis/blob/ecff13b29d938c8128dfe0cb51d22fd33c877c37/core/internal/state/load.go#L214

So, this commit restores the check by defining a dedicated error type
and wrapping it so that it can be caught further up the call stack.
```

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -26,15 +26,15 @@ import (
 	"time"
 
 	"github.com/krenalis/krenalis/cmd/internal/mcp"
-	"github.com/krenalis/krenalis/core"
+	corePkg "github.com/krenalis/krenalis/core"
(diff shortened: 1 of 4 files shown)


gazerro (commented):
Approved.

PR #2342, opened 2026-07-03
core: fix password reset race

gazerro:
```
core: fix password reset race

Previously SendMemberPasswordReset first looked up an active member by
email, then updated the password reset token in a separate query. A
concurrent change could therefore make the method write a reset token to
a row that was no longer an active member.

Now the reset token is written only when the matching row is still an
active member. If no row matches, SendMemberPasswordReset preserves the
existing behavior and does not reveal whether the email belongs to a
member or to a pending invitation.
```

### DIFF
--- a/core/organization.go
+++ b/core/organization.go
@@ -742,24 +742,15 @@ func (this *Organization) SendMemberPasswordReset(ctx context.Context, email str
 		return err
 	}
 	now := time.Now().UTC()
-	err = this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
-		result, err := tx.Exec(ctx, `UPDATE members SET reset_password_token = $1, reset_password_token_created_at = $2`+
-			` WHERE organization = $3 AND email = $4 AND invitation_token = ''`, resetToken, now, this.organization.ID, email)
-		if err != nil {
-			return nil, err
-		}
-		if result.RowsAffected() == 0 {
-			return nil, errMemberNotFoundOrInvitationPending
-		}
-		return nil, nil
-	})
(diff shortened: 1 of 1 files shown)


core/organization.go
```diff
@@ -743,16 +743,15 @@ func (this *Organization) SendMemberPasswordReset(ctx context.Context, email str
 	}
 	now := time.Now().UTC()
 	err = this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
-		exists, err := tx.QueryExists(ctx, "SELECT FROM members WHERE organization = $1 AND email = $2 AND invitation_token = ''", this.organization.ID, email)
+		result, err := tx.Exec(ctx, `UPDATE members SET reset_password_token = $1, reset_password_token_created_at = $2`+
```
zapateo: Here we should stop using the transaction.

zapateo (changes requested):

core/organization.go
gazerro: Done in commit [Remove unnecessary transaction](https://github.com/krenalis/krenalis/pull/2342/changes/87a78efb714a147b9f1c2734b1071a93ed183d84).

zapateo: Note that this error is no longer used, so it should be removed:

https://github.com/krenalis/krenalis/blob/87a78efb714a147b9f1c2734b1071a93ed183d84/core/organization.go#L724

zapateo (changes requested):

core/organization.go
gazerro: Done in commit [Remove unused error variable](https://github.com/krenalis/krenalis/pull/2342/changes/c5337a30342db42fa5ef5b3e92aca554a375acb4).

zapateo (approved):

PR #2349, opened 2026-07-05
cmd, core: enforce per-organization resource limits

gazerro:
```
cmd, core: enforce per-organization resource limits

Configure and enforce per-organization limits on members, access keys,
workspaces, connectors, connections, and pipelines.
```


gazerro:
See also krenalis/krenalis.com/pull/40.

core/organization.go
```diff
@@ -211,6 +239,17 @@ func (this *Organization) AddMember(ctx context.Context, member MemberToSet) (st
 		})
 		now := time.Now().UTC()
 		err = this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
+			// Ensure the organization can accept another member.
+			var count, limit int
+			err = tx.QueryRow(ctx, "SELECT (SELECT COUNT(*) FROM members m WHERE m.organization = o.id), o.members_limit"+
```
zapateo, line 244: In some places (like this one) _all members_ are counted (including those with an invitation token), while in other places (like `load.go`) _only members without an invitation token_ are counted.

This needs to be reviewed.

zapateo (changes requested):

core/organization.go
gazerro: Fixed in commit [Fix member usage count on state load](https://github.com/krenalis/krenalis/pull/2349/changes/4d2dedaa636768276d77160ff8c05166b6baf299).

core/internal/state/keep.go
zapateo, line 890: As discussed, we should rename the notification to something that clarifies that it's specific to WorkOS members, to emphasize that it doesn't _remove members_ of any kind (who can be active or invited), but only WorkOS members, who are always active by design.

This isn't related to this PR, so it should be done in `main` to remove the ambiguity we have at this point.

core/organization.go
```diff
@@ -588,6 +658,7 @@ func (this *Organization) HasMember(id string) (bool, error) {
 // It returns an errors.UnprocessableError error with code
 //   - EmailSendFailed, if emails cannot be sent.
 //   - MemberEmailExists, if the email address belongs to an existing member.
+//   - MembersLimitReached, if the organization cannot have more members.
 func (this *Organization) InviteMember(ctx context.Context, email string, emailTemplate string) error {
```
zapateo, line 658: This method does not send any notification, so the state is not informed when a member is added and its usage counter is never incremented.

core/internal/state/keep.go
```diff
@@ -326,6 +326,7 @@ func (state *State) acceptInvitation(n notification) string {
 	org := state.organizations[e.Organization]
 	org.mu.Lock()
 	org.members[e.Member] = struct{}{}
+	org.usage.addMember()
```
zapateo: Invited members are already counted, so here a member who accepts the invitation would be counted twice.

core/organization.go
zapateo: In case of an invited member being deleted, no notification is sent, so the member usage count would not be decremented in the state.

zapateo (changes requested):

core/organization.go
gazerro: Fixed in commit [Fix member usage state update on invitation](https://github.com/krenalis/krenalis/pull/2349/changes/5f039492404c4ec7e3815fbe42df89f52acdc213).

core/internal/state/keep.go
gazerro: Fixed in commit [Fix member usage state update on invitation](https://github.com/krenalis/krenalis/pull/2349/changes/5f039492404c4ec7e3815fbe42df89f52acdc213).

core/organization.go
gazerro: Fixed in commit [Fix member usage decrement when deleting invitations](https://github.com/krenalis/krenalis/pull/2349/changes/b98b86d9251b760883cf35c3c53d9aaf2612ad79).

core/internal/state/keep.go
```diff
@@ -104,6 +104,8 @@ func (state *State) keep() {
 			org = state.endIdentityResolution(n)
 		case "EndPipelineRun":
 			org = state.endPipelineRun(n)
+		case "InviteMember":
```
zapateo, line 107: `InviteMember` is a new notification type, so it should also be added to the db init script and in the updater, shouldn't it?

zapateo (changes requested):

core/internal/state/keep.go
gazerro: Fixed in commit [Add InviteMember notification to database schema](https://github.com/krenalis/krenalis/pull/2349/changes/1462f9b28b14246acf7a157d6c0d8a97cbce9cad).

zapateo (approved):
PR approved.

On krenalis.com, in documentation of fields referring to the "number of connectors", please clarify:

* that this is not the "number of connectors" itself, an unclear expression, but rather the _connectors used_.

* that this refers to unique connectors, not the actual number of connectors used.

* that this value takes into account both the connectors used by connections and those used as the format for connection file pipelines.

PR #2352, opened 2026-07-06
core/internal/state: clarify why some state reads do not require locking

gazerro:
```
core/internal/state: clarify why some state reads do not require locking

Only the goroutine running keep can write to the state, so reads
performed by that goroutine are safe without acquiring the corresponding
locks for the affected state parts.
```

### DIFF
--- a/core/internal/state/keep.go
+++ b/core/internal/state/keep.go
@@ -60,6 +60,8 @@ func (state *State) keep() {
 		}
 		var org string
 		state.changing.Lock()
+		// While state.changing is locked, only the current goroutine may modify the state.
+		// Other goroutines may only read it, so handlers do not need to acquire the lock for read-only access.
 		switch n.Name {
 		case "AcceptInvitation":
 			org = state.acceptInvitation(n)
--- a/core/internal/state/keep.go
+++ b/core/internal/state/keep.go
@@ -60,8 +60,9 @@ func (state *State) keep() {
 		}
 		var org string
 		state.changing.Lock()
-		// While state.changing is locked, only the current goroutine may modify the state.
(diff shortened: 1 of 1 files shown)


core/internal/state/keep.go
```diff
@@ -60,6 +60,8 @@ func (state *State) keep() {
 		}
 		var org string
 		state.changing.Lock()
+		// While state.changing is locked, only the current goroutine may modify the state.
```
zapateo: As discussed verbally, this comment should be replaced by something like:

```
// The state is read and written concurrently by multiple goroutines,
// but since the current goroutine is the only one that can write it,
// its read accesses do not require acquiring state locks.
```

core/internal/state/keep.go
zapateo, line 1: Remember to update the commit message. In particular:

* The commit message body should be updated (`Clarify that while state.changing is locked...`).
* The commit message title refers to a single lock (`needs no lock `), maybe there is a form that highlights that it refers to all state locks in general, not just a single, specific lock. I don't know if there's this ambiguity in English that I'm seeing, you decide.

zapateo (changes requested):

core/internal/state/keep.go
gazerro: Done in commit [Correct and clarify comment](https://github.com/krenalis/krenalis/pull/2352/changes/5887e3939f15225512477ad11f6bf42de8efe41d).

core/internal/state/keep.go
gazerro: Done.

zapateo (approved):

PR #2355, opened 2026-07-07
all: add consent management

retini:
Introduce consent management, allowing a workspace to declare the purposes for
which users may grant or deny consent, and to process events only when the
consents required by a pipeline are granted.

With these changes:

* Events accept a new 'context.consents' property, a map of purpose
  codes to booleans;

* A new Privacy page, under Settings, lets users add, edit and delete
  the consent purposes of a workspace;

* Pipelines and event listeners can require one or more consent
  purposes, combined with an "all" or "any" logical operator. Events
  that do not satisfy the required consents are discarded;

* The pipeline metrics gain a Consent step, reporting how many events passed the
  consent check and how many were discarded by it.


connectors/applications.go
```diff
@@ -590,6 +590,7 @@ type ReceivedEventContext interface {
 	App() (ReceivedEventContextApp, bool)
 	Browser() (ReceivedEventContextBrowser, bool)
 	Campaign() (ReceivedEventContextCampaign, bool)
+	Consent() (map[string]bool, bool)
```
gazerro: I would return an iterator and have it return all purposes.

`Consent() (iter.Seq2[string, bool], bool)`

This avoids having to allocate a new map.

gazerro: I think it would be better to call it `Consents`, since it represents multiple consents rather than a single one.

core/consent_purpose.go
```diff
@@ -0,0 +1,210 @@
...
+
+// CreateConsentPurpose creates a consent purpose with the given name and code,
+// and returns its identifier.
+//
+// name and code must be between 1 and 100 runes long.
+//
+// It returns an errors.UnprocessableError error with code
+// ConsentPurposeCodeExists if a consent purpose with the same code already
+// exists in the workspace.
+func (this *Workspace) CreateConsentPurpose(ctx context.Context, name, code string) (string, error) {
```
gazerro: The purpose already exists in a company, so it is being _added_ rather than _created_, just like members. The method should therefore be called `AddConsentPurpose`.

Rename it in one commit, then, in the following commit, only move the declaration so that it is alphabetically ordered within the file.

core/internal/state/state.go
```diff
@@ -794,6 +795,28 @@ func (workspace *Workspace) AccountByCode(connector, code string) (*Account, boo
 	return a, a != nil
 }
 
+// ConsentPurpose returns the consent purpose of the workspace with identifier
```
gazerro: Put them in alphabetical order in the file, after `Connections`.

core/internal/state/state.go
```diff
@@ -1507,40 +1547,42 @@ func (target Target) Value() (driver.Value, error) {
...
+	Target                  Target
+	Name                    string
+	Enabled                 bool
+	EventType               string
+	ScheduleStart           int16
+	SchedulePeriod          int16
+	InSchema                types.Type
+	OutSchema               types.Type
+	Filter                  *Where
+	RequiredConsents        []string
```
gazerro: `RequiredConsents` is not a list of required consents, but a list of purposes. A better name would be:

```go
type Pipeline struct {
    ...
    RequiredConsents RequiredConsents
    ...
}

type RequiredConsents struct {
    Operator ConsentPurposesOperator
    Purposes []string
}
```

core/internal/state/state.go
```diff
@@ -1707,3 +1749,13 @@ const (
 	UpdateOnly     ExportMode = "UpdateOnly"
 	CreateOrUpdate ExportMode = "CreateOrUpdate"
 )
+
+// RequiredConsentsLogical represents the logical operator applied to the
+// pipeline's required consents.
+type RequiredConsentsLogical string
```
gazerro: `ConsentsNone` is unnecessary. If there is only one purpose, both `and` and `or` work; if there are no purposes, `and` also works. Therefore, use `and` instead of `none`. Also, use integers for consistency with `Where`.

```go
type ConsentPurposesOperator int

const (
	PurposesAnd ConsentPurposesOperator = iota // and
	PurposesOr                                 // or
)
```

core/internal/state/state.go
```diff
@@ -1169,6 +1192,23 @@ func (account *Account) Connector() *Connector {
 	return c
 }
 
+// ConsentPurpose represents a consent purpose.
+type ConsentPurpose struct {
+	mu        *sync.Mutex
+	ID        string
+	workspace *Workspace
```
gazerro: This field is not used (or rather, it is used by the `Workspace` method which itself is not used.) I would therefore remove it.

core/consent_purpose.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/krenalis/krenalis/core/internal/util"
+	"github.com/krenalis/krenalis/tools/errors"
+)
+
+// ConsentPurpose represents a purpose.
+type ConsentPurpose struct {
+	ID        string                   `json:"id"`
+	Name      string                   `json:"name"`
+	Code      string                   `json:"code"`
+	Pipelines []ConsentPurposePipeline `json:"pipelines"`
```
gazerro: I think the pipelines that use a purpose are information relevant only to the Admin console and do not necessarily need to be exposed by the API at this point. The Admin console should determine this independently.

core/consent_purpose.go
```diff
@@ -0,0 +1,210 @@
...
+	Connector  string `json:"connector"`
+}
+
+// ConsentPurposes returns the consent purposes of the workspace, ordered by
+// name.
+func (this *Workspace) ConsentPurposes(ctx context.Context) ([]*ConsentPurpose, error) {
+	this.core.mustBeOpen()
+	purposes := make([]*ConsentPurpose, 0)
+	byID := make(map[string]*ConsentPurpose)
+	err := this.core.db.QueryScan(ctx,
```
gazerro: The purposes should be read from the state, since they are already available there.

core/consent_purpose.go
```diff
@@ -0,0 +1,210 @@
...
+	var id string
+	for {
+		id = generateID[any](nil)
+		err := this.core.state.Transaction(ctx, func(tx *db.Tx) (any, error) {
+			_, err := tx.Exec(ctx, "INSERT INTO consent_purposes (id, workspace, name, code) VALUES ($1, $2, $3, $4)",
+				id, ws.ID, name, code)
+			if err != nil {
+				return nil, err
+			}
+			return state.CreateConsentPurpose{ID: id, Workspace: ws.ID, Name: name, Code: code}, nil
```
gazerro: By convention, a variable named `n` containing the notification is declared before the transaction:

```go
n = state.CreateConsentPurpose{
    Workspace: this.workspace.ID,
    Name:      name,
    Code:      code,
}
```

Then, inside the transaction code, refer to the fields of `n`:

```go
n.ID = generateID[any](nil)
..
_, err := tx.Exec(..., n.ID, n.Workspace, n.Name, n.Code)
```

This creates a strong association between the database changes and the corresponding state changes.

Do the same for `UpdateConsentPurpose` and `DeleteConsentPurpose`.

core/consent_purpose.go
```diff
@@ -0,0 +1,210 @@
...
+	return id, nil
+}
+
+// UpdateConsentPurpose updates the name and code of the consent purpose with
+// identifier id.
+//
+// It returns an errors.NotFoundError error if the consent purpose does not
+// exist.
+//
+// It returns an errors.UnprocessableError error with code
```
gazerro, line 131: The error returned when the purpose does not exist has not been documented.

core/consent_purpose.go
```diff
@@ -0,0 +1,210 @@
...
+	if !IsValidID(id) {
+		return errors.BadRequest("identifier %q is not a valid consent purpose identifier", id)
+	}
+	if err := util.ValidateStringField("name", name, 100); err != nil {
+		return errors.BadRequest("%s", err)
+	}
+	if err := util.ValidateStringField("code", code, 100); err != nil {
+		return errors.BadRequest("%s", err)
+	}
+	ws := this.workspace
```
gazerro: First check in the state whether the purpose exists. This avoids starting a transaction.

core/pipeline_validate.go
```diff
@@ -164,6 +170,31 @@ func validatePipelineToSet(pipeline PipelineToSet, v validationState) error {
...
+	if len(pipeline.RequiredConsents) > 0 {
+		if !requiredConsentsAllowed {
+			return errors.BadRequest("required consents are not allowed")
+		}
+		seen := make(map[string]bool, len(pipeline.RequiredConsents))
+		for _, id := range pipeline.RequiredConsents {
+			if !IsValidID(id) {
+				return errors.BadRequest("identifier %q is not a valid consent purpose identifier", id)
+			}
+			if seen[id] {
```
gazerro: Use `slices.Contains(pipeline.RequiredConsents[i+1:], id)` as the condition. The code is simpler, does not allocate, and for such small slices, much more efficient.

core/pipeline_validate.go
```diff
@@ -164,6 +170,31 @@ func validatePipelineToSet(pipeline PipelineToSet, v validationState) error {
 			usedInPaths = properties
 		}
 	}
+	// Validate the required consents.
+	requiredConsentsAllowed := dispatchEventsToAplications || importEventsIntoWarehouse || importUserIdentitiesFromEvents
+	if len(pipeline.RequiredConsents) > 0 {
+		if !requiredConsentsAllowed {
+			return errors.BadRequest("required consents are not allowed")
+		}
+		seen := make(map[string]bool, len(pipeline.RequiredConsents))
```
gazerro: I would set a maximum limit on the length of `pipeline.RequiredConsents`. Use a very high number as a hard upper bound, for example `100`.

core/pipeline_validate.go
```diff
@@ -164,6 +170,31 @@ func validatePipelineToSet(pipeline PipelineToSet, v validationState) error {
...
+		for _, id := range pipeline.RequiredConsents {
+			if !IsValidID(id) {
+				return errors.BadRequest("identifier %q is not a valid consent purpose identifier", id)
+			}
+			if seen[id] {
+				return errors.BadRequest("required consent purpose %q is duplicated", id)
+			}
+			seen[id] = true
+			if !v.knownConsentPurposeIDs[id] {
+				return errors.Unprocessable(ConsentPurposeNotExist, "consent purpose %q does not exist", id)
```
gazerro: When adding a Base58 ID to an error message, use `%s` instead of `%q`. Use `%q` when displaying a string that was expected to be a valid ID but is not.

core/pipeline.go
```diff
@@ -641,6 +666,11 @@ func (this *Pipeline) Update(ctx context.Context, pipeline PipelineToSet) error
 
 	this.core.mustBeOpen()
 
+	// Normalize the required consents.
```
gazerro: Perform normalization after validation.

core/pipeline.go
```diff
@@ -1049,6 +1084,8 @@ func (this *Pipeline) fromState(core *Core, store *datastore.Store, pipeline *st
 	if pipeline.Filter != nil {
 		this.Filter = convertWhereToFilter(pipeline.Filter, pipeline.InSchema)
 	}
+	this.RequiredConsents = pipeline.RequiredConsents
```
gazerro: All values returned by the API must be copies of the data stored in `core`, so that callers cannot modify the original data, even accidentally:

```go
this.RequiredConsents = slices.Clone(pipeline.RequiredConsents)
```

core/internal/collector/decoder.go
```diff
@@ -857,6 +857,17 @@ func (d *decoder) decodeContext() (map[string]any, error) {
 				return nil, errors.BadRequest("property 'context.traits' is not a valid object")
 			}
 			context["traits"], _ = d.dec.ReadValue()
+		case "consent":
```
gazerro: `consents` is better than `consent`.

gazerro (changes requested):

connectors/applications.go
retini: Done in [`Make Consent return an iterator instead of a map`](https://github.com/krenalis/krenalis/pull/2355/changes/4320108310986cc736bc8d06bb5e5d4cbbb72fcb)

retini: Done in [`Rename Consent to Consents`](https://github.com/krenalis/krenalis/pull/2355/changes/51b4f42e92744cf736c1c1ddd74d3ba289059cf1)

core/consent_purpose.go
retini: Done in:

1. [`Rename CreateConsentPurpose to AddConsentPurpose`](https://github.com/krenalis/krenalis/pull/2355/changes/864bc2393a3a78e3c1662ca30d43bab1c1c5fc6a)
1. [`Update AddConsentPurpose and addConsentPurpose comments`](https://github.com/krenalis/krenalis/pull/2355/changes/8edd68fb650a2ed2658a1ed3f70b317409d7e480)
1. [`Alphabetically sort declarations related to adding consent purposes`](https://github.com/krenalis/krenalis/pull/2355/changes/d423834685e919de4321b416c128fcbf4cbcec34)

core/internal/state/state.go
retini: Done in [`Sort ConsentPurpose and ConsentPurposes in state.go`](https://github.com/krenalis/krenalis/pull/2355/changes/166dcaa439772a1f35ab1bc857d30833b7eeab05)

core/internal/state/state.go
retini: Done in [`Remove workspace field from ConsentPurpose type`](https://github.com/krenalis/krenalis/pull/2355/changes/c29baa2ac617c61c771671d6d501f736fa719d21)

core/internal/state/state.go
retini: Done in [`Group required consents into a RequiredConsents type`](https://github.com/krenalis/krenalis/pull/2355/changes/17b17b334f82e23d13cfae8a571127eaa70cc8df)

core/internal/state/state.go
retini: Done in [`Use and as the default consent purposes operator`](https://github.com/krenalis/krenalis/pull/2355/changes/95efb1d6cd624791f20cc8d6cea1edc0adb83ebe)

core/consent_purpose.go
retini: Done in [`Infer consent purpose pipelines in Admin console`](https://github.com/krenalis/krenalis/pull/2355/changes/2c820a37ca228145e0e20c7f06db7f6650ff4293)

core/consent_purpose.go
retini: Done in [`Read consent purposes from the state`](https://github.com/krenalis/krenalis/pull/2355/changes/9376e236f6462afb5d0dad7c1880cd752763ef37)

core/consent_purpose.go
retini: Done in [`Declare consent purpose notifications before transactions`](https://github.com/krenalis/krenalis/pull/2355/changes/02c2394f10a3eb9613d51ea5f1fe61fd903c4c1a)

core/consent_purpose.go
retini: Done in [`Check consent purpose existence in the state`](https://github.com/krenalis/krenalis/pull/2355/changes/110a44d31e0392b8b948e9b9fc0b988a1c088768)

core/pipeline.go
retini: Done in [`Return a copy of the required consent purposes`](https://github.com/krenalis/krenalis/pull/2355/changes/f18a8f2d56ef3b22dec8fde89a33e110ffc575ef)

core/pipeline.go
retini: Done in [`Perform normalization after validation`](https://github.com/krenalis/krenalis/pull/2355/changes/400f37984712876f056e075155a1f38abfb0fd99)

core/consent_purpose.go
retini: @gazerro It looks to me like that error is already documented. The comment explicitly says "It returns an errors.NotFoundError error if the consent purpose does not exist". Am I missing something?

core/pipeline_validate.go
retini: Done in [`Check duplicate consent purposes without a map`](https://github.com/krenalis/krenalis/pull/2355/changes/e4ab756f0c016802e08480eb8b9271930c3b04b2)

core/pipeline_validate.go
retini: Done in [`Set a maximum number of required consent purposes in the pipeline`](https://github.com/krenalis/krenalis/pull/2355/changes/e597fcae266cadbe545394dddbda9832468489e4)

core/pipeline_validate.go
retini: Done in [`Do not quote valid consent purpose IDs in error messages`](https://github.com/krenalis/krenalis/pull/2355/changes/d3c2beb476b8fd0bf5518a29250d16ef777f1862)

core/internal/collector/decoder.go
retini: Done in [`Rename context.consent to context.consents`](https://github.com/krenalis/krenalis/pull/2355/changes/d10c8599945cc894aee28a1c00d6bcb485767650)

core/internal/connections/received_event.go
```diff
@@ -119,15 +120,17 @@ func (c receivedEventContext) Campaign() (connectors.ReceivedEventContextCampaig
 	return nil, false
 }
 
-func (c receivedEventContext) Consent() (map[string]bool, bool) {
+func (c receivedEventContext) Consent() (iter.Seq2[string, bool], bool) {
 	if consent, ok := c.context["consent"].(map[string]any); ok {
```
gazerro: Return early; it is more idiomatic in Go.

```go
consent, ok := c.context["consent"].(map[string]any)
if !ok {
    return nil, false
}
...
```

core/internal/connections/received_event.go
```diff
@@ -119,15 +120,17 @@ func (c receivedEventContext) Campaign() (connectors.ReceivedEventContextCampaig
...
-		m := make(map[string]bool, len(consent))
-		for purpose, consented := range consent {
-			if v, ok := consented.(bool); ok {
-				m[purpose] = v
+		return func(yield func(string, bool) bool) {
+			for purpose, consented := range consent {
+				if v, ok := consented.(bool); ok {
```
gazerro: `consented` must be a boolean, as guaranteed by the decoder. It is better to panic if that is not the case.

```go
if !yield(purpose, consented.(bool)) {
     return
}
```

cmd/workspaces.go
```diff
@@ -85,6 +109,15 @@ func (workspace workspace) Connections(_ http.ResponseWriter, r *http.Request) (
 	return map[string]any{"connections": ws.Connections()}, nil
 }
 
+// ConsentPurposes returns the consent purposes of the current workspace.
+func (workspace workspace) ConsentPurposes(_ http.ResponseWriter, r *http.Request) (any, error) {
+	ws, err := workspace.workspace(r)
+	if err != nil {
+		return nil, err
+	}
+	return map[string][]*core.ConsentPurpose{"purposes": ws.ConsentPurposes()}, nil
```
gazerro: Here, `[]*core.ConsentPurpose` adds no value and only makes the code slightly less clear. See, for example, how this is handled in the previous `workspace.Connections` method.

```go
return map[string]any{"purposes": ws.ConsentPurposes()}, nil
```

cmd/workspaces.go
```diff
@@ -159,6 +193,19 @@ func (workspace workspace) Delete(_ http.ResponseWriter, r *http.Request) (any,
 	return nil, err
 }
 
+// DeleteConsentPurpose deletes a consent purpose of the current workspace.
+func (workspace workspace) DeleteConsentPurpose(_ http.ResponseWriter, r *http.Request) (any, error) {
+	if err := validateForbiddenBody(r); err != nil {
```
gazerro: This check has already been implemented; see PR #2377.

cmd/workspaces.go
```diff
@@ -123,9 +156,10 @@ func (workspace workspace) CreateEventListener(_ http.ResponseWriter, r *http.Re
 		return nil, err
 	}
 	var body struct {
-		Connection *string      `json:"connection"`
-		Size       *int         `json:"size"`
-		Filter     *core.Filter `json:"filter"`
+		Connection       *string               `json:"connection"`
+		Size             *int                  `json:"size"`
+		Filter           *core.Filter          `json:"filter"`
+		RequiredConsents core.RequiredConsents `json:"requiredConsents"`
```
gazerro: Use a pointer type.

core/internal/collector/decoder.go
```diff
@@ -1112,6 +1123,54 @@ func (d *decoder) decodeContextSection(section *contextSection) (map[string]any,
...
+		purpose := tok.String()
+		if purpose == "" {
+			return nil, errors.BadRequest("property 'context.consents' contains an empty purpose")
+		}
+		tok, err = d.dec.ReadToken()
+		if err != nil {
+			return nil, err
+		}
+		if tok.Kind() != json.True && tok.Kind() != json.False {
+			return nil, errors.BadRequest("property 'context.consents.%s' is not a valid boolean", purpose)
```
gazerro: `purpose` can contain any sequence of characters, so I would use `%q` instead of `%s`.

core/workspace.go
```diff
@@ -881,11 +896,25 @@ func (this *Workspace) CreateEventListener(connection string, size int, filter *
 		}
 		where = convertFilterToWhere(filter, schemas.Event)
 	}
+	consents := state.RequiredConsents{Operator: state.ConsentPurposesOperator(requiredConsents.Operator)}
```
gazerro: `requiredConsents` would be better than `consents`, since these aren't the consents themselves, but rather the data used to determine which consents are required.

core/workspace.go
```diff
@@ -881,11 +896,25 @@ func (this *Workspace) CreateEventListener(connection string, size int, filter *
...
+	if len(requiredConsents.Purposes) > 0 {
```
gazerro: If the purposes are converted from IDs to codes at this point, and those codes change while the observer is active, the observer will no longer enforce the consent checks correctly.

Also, using `state.RequiredConsents` when `Purposes` contains codes rather than IDs makes the code misleading, since readers would reasonably expect `state.RequiredConsents.Purposes` to contain purpose IDs.

core/internal/state/state.go
```diff
@@ -1707,3 +1739,57 @@ const (
 	UpdateOnly     ExportMode = "UpdateOnly"
 	CreateOrUpdate ExportMode = "CreateOrUpdate"
 )
+
+// RequiredConsents represents the consent purposes required by a pipeline.
+type RequiredConsents struct {
+	Operator ConsentPurposesOperator
+	Purposes []string
```
gazerro: For the current use cases, we need to know which purposes are associated with a pipeline and the codes of those purposes, with all the information referring to the same state version.

Suppose the following happens in this order:

1. I read a pipeline from the state, and its required purposes are `A and B and C`.
2. Purpose `A` is removed from the pipeline and purpose `D` is added, so the required purposes become `B and C and D`.
3. Purpose `A` is deleted.
4. I try to read purpose `A` from the state, but it no longer exists.

At this point, I have two possible options:

a) Treat the validation as failed. However, this would be incorrect, because the validation would have passed in both actual states: `A and B and C` and `B and C and D`.

b) Pretend that `A` was never present and treat the required purposes as `B and C`. This would also be incorrect, because there has never been a state in which the required purposes were `B and C`.

Therefore, the only correct approach is to ensure that all the information used to determine whether validation passes or fails refers to the same state version.

There are two ways to achieve this:

a) Freeze the state and read all the required information while it is frozen.

b) Have the `state` package provide a way to read all the required information together, while guaranteeing that it comes from the same consistent state version.

Option **a)** is currently used only when Krenalis starts, allowing the various subsystems to read a consistent state that is then kept synchronized with the `state` package. Therefore, I would implement option **b)**.

## Solution 1

The first solution is to change:

```go
Purposes []string
```

to:

```go
Purposes []*Purpose
```

This requires every modification to a purpose to be propagated to every pipeline that uses it. Those pipelines must then be replaced everywhere they are referenced.

## Solution 2

The second solution is to make this field and its related type:

```go
RequiredConsents RequiredConsents
```

unexported:

```go
requiredConsents requiredConsents
```

and add a new method and a new type:

```go
func (p *Pipeline) RequiredConsents() RequiredConsents

type RequiredConsents struct {
    Operator ConsentPurposesOperator
    Purposes []*Purpose
}
```

The downside is that each call to `Pipeline.RequiredConsents` requires allocating a new slice for `RequiredConsents.Purposes`.

Given how frequently we expect to call `Pipeline.RequiredConsents()`, I think the first solution is the better option.

core/consent_purpose.go
```diff
@@ -0,0 +1,185 @@
...
+	if err := util.ValidateStringField("name", name, 100); err != nil {
+		return errors.BadRequest("%s", err)
+	}
+	if err := util.ValidateStringField("code", code, 100); err != nil {
+		return errors.BadRequest("%s", err)
+	}
+	if _, ok := this.workspace.ConsentPurpose(id); !ok {
+		return errors.NotFound("consent purpose %s does not exist", id)
+	}
+	n := state.UpdateConsentPurpose{
```
gazerro, line 152: If `state.Pipeline.RequiredConsents.Purposes` is changed from `[]string` to `[]*Purpose`, it may be worth avoiding an update when its value has not changed.

```go
current, ok := this.workspace.ConsentPurpose(id)
if !ok {
    return errors.NotFound("consent purpose %s does not exist", id)
}
if name == current.Name && code == current.Code {
    return nil
} 
```

core/pipeline.go
```diff
@@ -853,7 +924,7 @@ func (this *Pipeline) Update(ctx context.Context, pipeline PipelineToSet) error
 		}
 		// Update the pipeline.
 		result, err := tx.Exec(ctx, update,
-			n.Name, n.Enabled, rawInSchema, rawOutSchema, n.Filter, mapping,
+			n.Name, n.Enabled, rawInSchema, rawOutSchema, n.Filter, n.RequiredConsents.Purposes, n.RequiredConsents.Operator, mapping,
```
gazerro, line 930: Before updating the pipeline, we must verify that all specified purposes exist in the workspace.

core/workspace.go
gazerro: I would also verify that there are no more than 100 purposes, as we already do for pipelines.

core/workspace.go
```diff
@@ -881,11 +896,25 @@ func (this *Workspace) CreateEventListener(connection string, size int, filter *
...
+		consents.Purposes = make([]string, len(requiredConsents.Purposes))
+		for i, id := range requiredConsents.Purposes {
+			if !IsValidID(id) {
+				return "", errors.BadRequest("identifier %q is not a valid consent purpose identifier", id)
+			}
+			cp, ok := this.workspace.ConsentPurpose(id)
```
gazerro: I would also check for duplicates, as we already do for pipelines.

gazerro (changes requested):

core/internal/connections/received_event.go
retini: Done in [`Return early in Consents()`](https://github.com/krenalis/krenalis/pull/2355/changes/6279b29f2efb0282f03dbc161bd2bcc1b86f2fda)

core/internal/connections/received_event.go
retini: Done in [`Panic on a non-boolean consent in Consents()`](https://github.com/krenalis/krenalis/pull/2355/changes/0cf3819de57d05f25dcfa3eb088e3b646104c616)

cmd/workspaces.go
retini: Done in [`Return a map[string]any from ConsentPurposes()`](https://github.com/krenalis/krenalis/pull/2355/changes/55cdf39abb0ceed05feff5f0759245261f2ddd8d)

cmd/workspaces.go
retini: Done in [`Remove the duplicate body validation in DeleteConsentPurpose()`](https://github.com/krenalis/krenalis/pull/2355/changes/004b80f80d1e8890f789db42a62f4e76662e6563)

cmd/workspaces.go
retini: Done in [`Make requiredConsents a pointer in CreateEventListener()`](https://github.com/krenalis/krenalis/pull/2355/changes/5b8e0102d74f145f30667bd19c5f5f7d7186f4a1)

core/internal/collector/decoder.go
retini: Done in [`Quote the purpose in the context.consents error message`](https://github.com/krenalis/krenalis/pull/2355/changes/6143c773d1615cdccb3347316ac1e0aa3c9940ef)

core/workspace.go
retini: Done in [`Rename consents to rc`](https://github.com/krenalis/krenalis/pull/2355/changes/a407ce3125b230eec0d4a87072e02670e2afde65).

I couldn't use `requiredConsents` though, since that's already the parameter holding the `*RequiredConsents`, so I went with `rc`, since the variable has a short scope.

core/workspace.go
retini: Done in [`Limit the number of required consent purposes in CreateEventListener()`](https://github.com/krenalis/krenalis/pull/2355/changes/87c54907d5401837a472f330e746ba7bde2893e7)

core/workspace.go
retini: Done in [`Check for duplicate required consent purposes in CreateEventListener()`](https://github.com/krenalis/krenalis/pull/2355/changes/e627187d8fba71070f2d89c5c5df20381f490512)

core/pipeline.go
retini: Done in [`Check existing required consent purposes before updating the pipeline`](https://github.com/krenalis/krenalis/pull/2355/changes/cd6f01b53b628e9b0aa7452256b9d535949862ad)

core/internal/state/state.go
retini: Done in [`Refer to the required consent purposes of a pipeline by pointer`](https://github.com/krenalis/krenalis/pull/2355/changes/7dc779da8caa52edc1fecdd8edc31fdb55cfd1f6)

core/consent_purpose.go
retini: Done in [`Skip the update if the consent purpose is unchanged`](https://github.com/krenalis/krenalis/pull/2355/changes/8319467568bff55ea9fa5f478351252279b81aff)

core/workspace.go
retini: Done in [`Replace an updated consent purpose in the event listeners`](https://github.com/krenalis/krenalis/pull/2355/changes/98d967ade7e6a4fb831ab458ce2384fea45131ef)

core/pipeline.go
gazerro: Within the transaction, we should check the state directly in the database rather than relying on the in-memory state, so we’ll need to run a query.

core/internal/state/state.go
```diff
@@ -839,6 +839,22 @@ func (workspace *Workspace) ConsentPurposes() []*ConsentPurpose {
 	return purposes
 }
 
+// resolveRequiredConsents returns the required consents of a pipeline of the
+// workspace, given the required consent purposes referred to by identifier.
+func (workspace *Workspace) resolveRequiredConsents(ids RequiredConsentIDs) RequiredConsents {
```
gazerro: This method is called only during initial loading or while the keeper is updating the state. In both cases, `workspace.consentPurposes` cannot be modified concurrently, so there is no need to acquire `workspace.mu`.

Acquiring the lock here could also cause a deadlock if the caller already holds `workspace.mu`. This can happen because the method is used as part of the state update path. Synchronization must therefore be handled by the caller, and this precondition should be documented explicitly.

Simply add the following line at the end of the method comment:

`// It must be called on a frozen state.`

gazerro: I'd try temporarily inlining its implementation at the call sites to see whether a simpler function with a clearer responsibility emerges.

core/internal/state/state.go
```diff
@@ -1743,7 +1759,16 @@ const (
...
 	Operator ConsentPurposesOperator
-	Purposes []string
+	Purposes []*ConsentPurpose
+}
+
+// RequiredConsentIDs represents the consent purposes required by a pipeline,
+// referred to by their identifiers. It is used where the consent purposes
+// cannot be referred to by pointer, as in the notifications that create and
+// update a pipeline and when reading a pipeline from the database.
+type RequiredConsentIDs struct {
```
gazerro: The name suggests that it represents IDs, but that is not actually the case. What do you think about renaming it to `RequiredConsentsByID`?

core/internal/state/keep.go
```diff
@@ -1649,10 +1650,28 @@ func (state *State) updateConsentPurpose(n notification) string {
...
-	ws.replaceConsentPurpose(e.ID, func(cp *ConsentPurpose) {
+	old := ws.consentPurposes[e.ID]
+	cp := ws.replaceConsentPurpose(e.ID, func(cp *ConsentPurpose) {
 		cp.Name = e.Name
 		cp.Code = e.Code
 	})
+	// Replace the consent purpose in the pipelines that require it.
+	for _, c := range ws.connections {
+		for _, p := range c.pipelines {
+			if !slices.Contains(p.RequiredConsents.Purposes, old) {
```
gazerro: This is better:

```go
i := slices.Index(p.RequiredConsents.Purposes, old)
if i < 0 {
      continue
}
purposes := slices.Clone(p.RequiredConsents.Purposes)
purposes[i] = cp
state.replacePipeline(p.ID, func(p *Pipeline) {
     p.RequiredConsents.Purposes = purposes
})
```

It is simpler, more efficient, and keeps the pipeline locked for less time.

core/internal/state/keep.go
```diff
@@ -1772,6 +1791,8 @@ func (state *State) updatePipeline(n notification) string {
 		filter, _ = unmarshalWhere(e.Filter, e.InSchema)
 	}
 	oldFormat := state.pipelines[e.ID].format
```
gazerro, line 1768: This is better:

```go
old := state.pipelines[e.ID]
ws := old.connection.workspace

```

and then use `old.Format` instead of `oldFormat`.

core/workspace.go
gazerro: As discussed, this solution does not address the problem. As an alternative, I would add the following field to `State.Workspace`:

```go
PurposeIDToCode map[string]string // Maps each purpose ID to its code.
```

Because the field is exported, the state must guarantee that the map is immutable. Therefore, whenever a purpose is added or removed, or a purpose code is changed, the map must be cloned and updated, and the workspace must then be replaced using `replaceWorkspace`.

gazerro (changes requested):

core/pipeline.go
retini: Done in [`Check the pipeline purposes existence against the database`](https://github.com/krenalis/krenalis/pull/2355/changes/c37514193e077c56bc4b315690c6d3b80f8dc546)

core/workspace.go
retini: @gazerro as discussed, we now identify consent purposes by their code instead of by an ID, so this no longer applies. See [`Identify the consent purposes by code instead of by identifier`](https://github.com/krenalis/krenalis/pull/2355/changes/8f4044fab62f6f0dbe35917ec320b1b4475361e6).

core/internal/state/state.go
retini: @gazerro as discussed, we now identify consent purposes by their code instead of by an ID, so this no longer applies. See [`Identify the consent purposes by code instead of by identifier`](https://github.com/krenalis/krenalis/pull/2355/changes/8f4044fab62f6f0dbe35917ec320b1b4475361e6).

retini: @gazerro as discussed, we now identify consent purposes by their code instead of by an ID, so this no longer applies. See [`Identify the consent purposes by code instead of by identifier`](https://github.com/krenalis/krenalis/pull/2355/changes/8f4044fab62f6f0dbe35917ec320b1b4475361e6).

core/internal/state/state.go
retini: @gazerro as discussed, we now identify consent purposes by their code instead of by an ID, so this no longer applies. See [`Identify the consent purposes by code instead of by identifier`](https://github.com/krenalis/krenalis/pull/2355/changes/8f4044fab62f6f0dbe35917ec320b1b4475361e6).

core/internal/state/keep.go
retini: @gazerro as discussed, we now identify consent purposes by their code instead of by an ID, so this no longer applies. See [`Identify the consent purposes by code instead of by identifier`](https://github.com/krenalis/krenalis/pull/2355/changes/8f4044fab62f6f0dbe35917ec320b1b4475361e6).

core/internal/state/keep.go
retini: @gazerro as discussed, we now identify consent purposes by their code instead of by an ID, so this no longer applies. See [`Identify the consent purposes by code instead of by identifier`](https://github.com/krenalis/krenalis/pull/2355/changes/8f4044fab62f6f0dbe35917ec320b1b4475361e6).

cmd/workspaces.go
```diff
@@ -22,6 +22,27 @@ type workspace struct {
...
+	if err := validateRequiredBody(r, false); err != nil {
+		return nil, err
+	}
+	ws, err := workspace.workspace(r)
+	if err != nil {
+		return nil, err
+	}
+	var body struct {
+		Name string `json:"name"`
+		Code string `json:"code"`
```
gazerro, line 35: Now that the code uniquely identifies a purpose, place it before the name and apply the same ordering everywhere, including function parameters, struct changes, validation logic, table columns, and similar cases.

core/internal/state/keep.go
```diff
@@ -1563,6 +1618,34 @@ func (state *State) updateConnection(n notification) string {
 	return c.organization.ID
 }
 
+// UpdateConsentPurpose is the event sent when a consent purpose is updated.
+type UpdateConsentPurpose struct {
```
gazerro, line 1628:
```go
type UpdateConsentPurpose struct {
	Workspace string
	Purpose   string
	Code      string
	Name      string
}
```

core/internal/collector/decoder.go
```diff
@@ -1112,6 +1123,54 @@ func (d *decoder) decodeContextSection(section *contextSection) (map[string]any,
...
+		purpose := tok.String()
+		if purpose == "" {
+			return nil, errors.BadRequest("property 'context.consents' contains an empty purpose")
+		}
+		tok, err = d.dec.ReadToken()
+		if err != nil {
+			return nil, err
+		}
+		if tok.Kind() != json.True && tok.Kind() != json.False {
+			return nil, errors.BadRequest("property 'context.consents.%q' is not a valid boolean", purpose)
```
gazerro: The error message does sound a bit awkward. I'd phrase it like this:

```go
"value of %q in 'context.consents' is not a valid boolean"
```

core/internal/collector/observer.go
```diff
@@ -64,13 +65,14 @@ type Observer struct {
...
-	connections []string
-	filter      *state.Where
-	sync.Mutex  // for the events and omitted fields
-	events      []json.Value
-	times       []time.Time
-	omitted     int
+	id               string
+	connections      []string
+	filter           *state.Where
+	requiredConsents state.RequiredConsents
```
gazerro: I'd make it a pointer, as that makes it much clearer that the value may be absent and is also consistent with `filter`.

core/internal/state/state.go
```diff
@@ -816,6 +817,28 @@ func (workspace *Workspace) Connections() []*Connection {
...
+func (workspace *Workspace) ConsentPurposes() []*ConsentPurpose {
+	workspace.mu.Lock()
+	purposes := make([]*ConsentPurpose, len(workspace.consentPurposes))
+	i := 0
+	for _, cp := range workspace.consentPurposes {
+		purposes[i] = cp
+		i++
+	}
+	workspace.mu.Unlock()
+	return purposes
```
gazerro, line 873: Sort them by code before returning them:

```go
sort.Slice(...)
```

Sorting makes the code more deterministic: errors consistently report the elements in the same order, and the debugger always shows the slice with a stable ordering.

core/internal/state/state.go
```diff
@@ -1169,6 +1192,13 @@ func (account *Account) Connector() *Connector {
 	return c
 }
 
+// ConsentPurpose represents a consent purpose.
+type ConsentPurpose struct {
+	mu   *sync.Mutex
```
gazerro: The mutex is never used, which is also evident from the fact that the struct has no unexported fields.

core/connection.go
```diff
@@ -315,6 +315,7 @@ func (this *Connection) ApplicationUsers(ctx context.Context, schema types.Type,
 //
 //   - ConnectionNotExist, if the connection does not exist.
 //   - ConnectorsLimitReached, if the organization cannot have more connectors.
+//   - ConsentPurposeNotExist, if a required consent purpose does not exist, only when the pipeline is on events.
```
gazerro: I would remove "required" because it suggests that this method requires it. Also, “consent purpose” is not ambiguous on its own, since purposes are stored in a single field of `PipelineToSet`.

core/connection.go
```diff
@@ -402,6 +408,12 @@ func (this *Connection) CreatePipeline(ctx context.Context, target Target, event
 		if err != nil {
 			return "", errors.Unprocessable(SchemaNotAligned, "output schema is not aligned with the event type schema: %w", err)
 		}
+		// Check that the required consent purposes exist, so that the events do
```
gazerro: Schema alignment is validated here for pipelines that send events to a destination because, otherwise, the check would have to be performed when Krenalis starts. If the schema of even one of those pipelines were misaligned, Krenalis would have no choice but to fail startup with an error. See: https://github.com/krenalis/krenalis/commit/444c4db9b71df461f0020bdfcc512ce18e1c8811

The issue is therefore not simply that events would no longer be processed by that pipeline. The impact is much more severe: Krenalis would fail to start entirely.

By contrast, if a required consent purpose does not exist, the worst-case outcome is that events for that pipeline fail at the Consent step.

For this reason, this validation should not be performed here and when a pipeline is updated.

core/internal/initdb/schema.sql
```diff
@@ -93,6 +93,13 @@ CREATE TABLE access_keys (
     PRIMARY KEY (id)
 );
 
+CREATE TABLE consent_purposes (
+    workspace varchar(12) NOT NULL REFERENCES workspaces ON DELETE CASCADE,
+    code varchar(100) NOT NULL CHECK (code ~ '^[0-9A-Za-z._-]{1,100}$'),
```
gazerro: This is not the format used for properties.

In this format, `.` and `..` are valid values. However, if they are included in a URL path without being escaped, they retain their special path semantics. For example, `/consent-purposes/..` would be resolved to `/`.

Using the properties format introduces a different issue: numeric values are not allowed. Therefore, if a CMP code consists only of digits, it cannot be represented as a purpose code. Once aliases are supported, however, the best solution will be to use an alias, especially since a code such as `2386` provides no indication of what the purpose actually is.

core/consent_purpose.go
```diff
@@ -0,0 +1,189 @@
...
+
+// ConsentPurpose represents a purpose.
+type ConsentPurpose struct {
+	Name string `json:"name"`
+	Code string `json:"code"`
+}
+
+// AddConsentPurpose adds a consent purpose with the given name and code.
+//
+// name must be between 1 and 100 runes long and code must be a valid consent
```
gazerro: `must be a valid consent purpose code`: the correct format for a purpose code is not documented anywhere.

Unlike a connection or pipeline ID, which is never chosen by the caller, a purpose code is provided by the caller. The caller therefore needs to know which format is considered valid.

There are two possible solutions:

* Document the format in every method that accepts a caller-defined code: `AddConsentPurpose`, `SetConsentPurpose`, and `CreateEventListener`. This would not be necessary for methods such as `CreatePipeline`, where the caller is expected to pass a code that was previously retrieved from the API.

* Add an exported `IsValidConsentPurposeCode` function that also documents the expected format. The API methods could then use this function internally to validate purpose codes.

I recommend the second approach.

core/workspace.go
```diff
@@ -881,11 +890,30 @@ func (this *Workspace) CreateEventListener(connection string, size int, filter *
 		}
 		where = convertFilterToWhere(filter, schemas.Event)
 	}
+	var rc state.RequiredConsents
+	if requiredConsents != nil {
+		if op := requiredConsents.Operator; op != PurposesAnd && op != PurposesOr {
+			return "", errors.BadRequest(`required consents operator must be "and" or "or"`)
+		}
+		if len(requiredConsents.Purposes) > MaxRequiredConsentPurposes {
+			return "", errors.BadRequest("required consent purposes must be at most %d", MaxRequiredConsentPurposes)
```
gazerro, line 766: The method documentation must specify this.

core/workspace.go
```diff
@@ -881,11 +890,30 @@ func (this *Workspace) CreateEventListener(connection string, size int, filter *
...
+		}
+		rc.Operator = state.ConsentPurposesOperator(requiredConsents.Operator)
+		rc.Purposes = slices.Clone(requiredConsents.Purposes)
+		for i, code := range rc.Purposes {
+			if !consentPurposeCodeFormat.MatchString(code) {
+				return "", errors.BadRequest("new code must be between 1 and 100 characters long and can only contain letters, digits, dots, hyphens and underscores")
```
gazerro: `new code` is incorrect because the code is not new, while `code` on its own does not make it clear what it refers to. The actual value is also missing. A better message would be: `code %q is not a valid consent purpose code`.

Also, `can only contain letters, digits, dots, hyphens, and underscores` may be ambiguous when shown to an end user (for example, `à` is a letter, but it is not allowed.) However, since this message will most often be read by a developer, a more technical format would be appropriate, see this example:

https://github.com/krenalis/krenalis/blob/a2a4cb9a2a0f0d8af2457a6f44e84dc2f1b02a2e/core/internal/connections/connections.go#L565-L566

Review the other error messages using the same criteria.

core/workspace.go
```diff
@@ -881,11 +890,30 @@ func (this *Workspace) CreateEventListener(connection string, size int, filter *
...
+			}
+			if slices.Contains(rc.Purposes[i+1:], code) {
+				return "", errors.BadRequest("required consent purpose %q is duplicated", code)
```
gazerro, line 777: The method documentation must specify this.

core/pipeline.go
gazerro: The code has since changed, so this modification is no longer relevant.

core/consent_purpose.go
```diff
@@ -0,0 +1,189 @@
...
+// UpdateConsentPurpose updates the consent purpose with the given code, setting
+// its name to name and its code to newCode.
+//
+// It returns an errors.NotFoundError error if the consent purpose does not
+// exist.
+//
+// It returns an errors.UnprocessableError error with code
+// ConsentPurposeCodeExists if another consent purpose with code newCode already
+// exists in the workspace.
+func (this *Workspace) UpdateConsentPurpose(ctx context.Context, code, name, newCode string) error {
```
gazerro: Better:

```go
func (this *Workspace) UpdateConsentPurpose(ctx context.Context, code string, purpose *ConsentPurpose) error {
```

This avoids introducing a conflict between `code` and `newCode`, and the method signature will not need to change when new fields are added, as we plan to do.

core/consent_purpose.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		if db.IsUniqueViolation(err) && db.ErrConstraintName(err) == "consent_purposes_pkey" {
+			return errors.Unprocessable(ConsentPurposeCodeExists, "a consent purpose with code %q already exists", n.NewCode)
+		}
+		return err
+	}
+	return nil
+}
+
+// DeleteConsentPurpose deletes the consent purpose with the given code.
```
gazerro, line 84: Sort the methods alphabetically within the file.

core/pipeline.go
```diff
@@ -682,6 +745,17 @@ func (this *Pipeline) Update(ctx context.Context, pipeline PipelineToSet) error
 		if err != nil {
 			return errors.Unprocessable(SchemaNotAligned, "output schema is not aligned with the event type schema: %w", err)
 		}
+		// Check that the required consent purposes exist, so that the events do
```
gazerro: Remove this check, as mentioned in the other comment, and move the previous comment back to its original position.

core/internal/connections/received_events_test.go
```diff
@@ -92,6 +93,11 @@ func Test_ReceivedEvent(t *testing.T) {
 				"id":    1751031467043,
 				"start": true,
 			},
+			"consents": map[string]any{
```
gazerro: Move it between `campaign` and `device`, as that is its natural position. Do the same in the other places within this function.

gazerro (changes requested):

cmd/workspaces.go
retini: Done in [`Reorder the consent purpose code and name`](https://github.com/krenalis/krenalis/pull/2355/changes/75e28c209043c366d55bdce105fad9c61815e66f)

core/internal/state/keep.go
retini: Done in:
1. [`Rename the parameters of the consent purpose update`](https://github.com/krenalis/krenalis/pull/2355/changes/8071222fa84b225e01a696fe4480a4e74ef01857)
2. [`Fix stale newCode reference in UpdateConsentPurpose documentation`](https://github.com/krenalis/krenalis/pull/2355/changes/2d95a79e2876527e643edc018b8925915803ca36)

core/internal/collector/decoder.go
retini: Done in:

1. [`Update error message in decodeConsents`](https://github.com/krenalis/krenalis/pull/2355/changes/0e71e25d72e98129340d97e6f3261440ee0d9147)
2. [`Align the decoder test with the new decodeConsents error message`](https://github.com/krenalis/krenalis/pull/2355/changes/9852a95afae3b33c90778a4a1c0f95c2b866286f)

core/internal/collector/observer.go
retini: Done in [`Make the required consents of an event listener a pointer`](https://github.com/krenalis/krenalis/pull/2355/changes/e27dd85292fdb8f86a02ef05f8e7a4acef684579)

core/internal/state/state.go
retini: Done in [`Sort the consent purposes returned by the state by code`](https://github.com/krenalis/krenalis/pull/2355/changes/a7ed345a8e4f53304e9b9fff026868971717242a)

core/internal/state/state.go
retini: Done in [`Remove the unused mutex from the state consent purpose`](https://github.com/krenalis/krenalis/pull/2355/changes/cc4f8b3f34bff9651672c85917f90ba7080ec164)

core/connection.go
retini: Done in [`Simplify ConsentPurposeNotExist error documentation in CreatePipeline`](https://github.com/krenalis/krenalis/pull/2355/changes/b506a91fae5806b4146dde36796f2752296181bd)

core/connection.go
retini: Done in [`Do not check the required consent purposes existence on pipelines`](https://github.com/krenalis/krenalis/pull/2355/changes/3dc1e964783a3bad34938d9ebe5ac65ffd51112b)

core/workspace.go
retini: Done in [`Document the requiredConsents constraints in CreateEventListener`](https://github.com/krenalis/krenalis/pull/2355/changes/ee5a295f5be5a5ab95eacfbee90f25bda0380c45)

core/workspace.go
retini: Done in [`Document the requiredConsents constraints in CreateEventListener`](https://github.com/krenalis/krenalis/pull/2355/changes/ee5a295f5be5a5ab95eacfbee90f25bda0380c45)

core/internal/initdb/schema.sql
retini: Done in [`Restrict the consent purpose code format to that of a property name`](https://github.com/krenalis/krenalis/pull/2355/changes/c69c4a9d935e65f8875dbaf9d38b674d862d9b94)

core/consent_purpose.go
retini: Commit [`Restrict the consent purpose code format to that of a property name`](https://github.com/krenalis/krenalis/pull/2355/changes/c69c4a9d935e65f8875dbaf9d38b674d862d9b94) already added the documentation of the correct format of a purpose code by extracting the validation into a `validateConsentPurposeCode` function.

Commit [`Add exported IsValidConsentPurposeCode function to document valid purpose codes`](https://github.com/krenalis/krenalis/pull/2355/changes/387439d5f0e30e39df6de8ae0961a38a8df26eb1) moves the documentation to a new `IsValidConsentPurposeCode` exported function, while preserving `validateConsentPurposeCode` for internal validation.

core/workspace.go
retini: Commit [`Restrict the consent purpose code format to that of a property name`](https://github.com/krenalis/krenalis/pull/2355/changes/c69c4a9d935e65f8875dbaf9d38b674d862d9b94) already updated the error message related to the consent purpose code validation.

Commits [`Clarify the consent purpose validation error messages`](https://github.com/krenalis/krenalis/pull/2355/changes/76cc101638ff69643a3673834261023503be7cba) and [`Report the invalid operator value in the required consents errors`](https://github.com/krenalis/krenalis/pull/2355/changes/ce5442ba17d4f210dfbf96ec136626c462db7e8e) review and improve the error messages as suggested by your comment.

core/consent_purpose.go
retini: Done in [`Pass ConsentPurpose to UpdateConsentPurpose instead of code and name`](https://github.com/krenalis/krenalis/pull/2355/changes/92ace8f04a3b69e4fc5b7a7d3bab6f918920a1cf)

core/consent_purpose.go
retini: Done in [`Sort the consent purposes methods alphabetically`](https://github.com/krenalis/krenalis/pull/2355/changes/1d57e679016635e60e6bf148cb9eaa3250dc76e4)

core/pipeline.go
retini: The check has already been removed by commit [`Do not check the required consent purposes existence on pipelines`](https://github.com/krenalis/krenalis/pull/2355/changes/3dc1e964783a3bad34938d9ebe5ac65ffd51112b).

Commit [`Restore comment position in pipeline Update`](https://github.com/krenalis/krenalis/pull/2355/changes/8aeae836ce5284ae16d218cec3a0b0e59388612a) restores the position of the comment.

core/internal/connections/received_events_test.go
retini: Done in [`Sort the consents context alphabetically in Test_ReceivedEvent`](https://github.com/krenalis/krenalis/pull/2355/changes/4a8160eac2f1c3be66543a1c30504caf7dfd6f66)

core/consent_purpose.go
```diff
@@ -173,3 +173,16 @@ func (this *Workspace) DeleteConsentPurpose(ctx context.Context, code string) er
...
 	})
 }
+
+// validateConsentPurposeCode validates the given consent purpose code.
+//
+// A valid consent purpose code has the same format as a property name and is at
+// most maxConsentPurposeCodeLen characters long.
+func validateConsentPurposeCode(code string) error {
+	if len(code) > maxConsentPurposeCodeLen || !types.IsValidPropertyName(code) {
+		return errors.BadRequest("code %q is not a valid consent purpose code. Consent purpose codes must be from 1 to %d"+
```
gazerro: `return fmt.Errorf(...)`

and at the call sites, use:

`return errors.BadRequest("%s", err)`

if `errors.BadRequestError` is indeed the error you intend to return.

gazerro (changes requested):

admin/src/lib/api/types/profile.ts
```diff
@@ -84,6 +84,7 @@ interface ProfileEventContext {
 	app?: ProfileEventContextApp;
 	browser?: ProfileEventContextBrowser;
 	campaign?: ProfileEventContextCampaign;
+	consent?: Record<string, boolean>;
```
gazerro: It must be `consents`.

admin/src/components/routes/PipelineWrapper/PipelineConsents.tsx
```diff
@@ -0,0 +1,131 @@
...
+			title='Privacy'
+			description='Choose whether this pipeline should require consent for specific purposes before processing events.'
+			padded={true}
+			ref={ref}
+			annotated={true}
+		>
+			<div className='pipeline__consents-toggle'>
+				<SlCheckbox checked={isEnabled} onSlChange={onToggle} disabled={purposes.length === 0} />
+				<div className='pipeline__consents-logical-sentence'>
+					An event must have consent for
```
gazerro, line 102: Could you make the entire checkbox label clickable, so clicking anywhere on it checks or unchecks the checkbox? The select element should, of course, remain excluded from this behavior.

core/consent_purpose.go
retini: Done in [`Return a plain error from validateConsentPurposeCode`](https://github.com/krenalis/krenalis/pull/2355/changes/d1d11c07a4c8accc47787ed9b33129e4409dd072)

admin/src/components/routes/Privacy/Privacy.tsx
```diff
@@ -0,0 +1,385 @@
...
+					<div className='privacy__dialog-error'>
+						<SlIcon slot='icon' name='exclamation-octagon' />
+						{codeError}
+					</div>
+				)}
+				{isEditing && code !== purposeToEdit.code && pipelines.length > 0 && (
+					<div className='privacy__dialog-warning'>
+						<SlIcon slot='icon' name='exclamation-triangle' />
+						{pipelines.length === 1
+							? `1 pipeline requires the "${purposeToEdit.code}" code. It will stop processing events until you update it too.`
```
gazerro: As discussed, we'll use the following text:

> This purpose is used by 1 pipeline. Changing its code may cause events to be filtered incorrectly.

gazerro (changes requested):

admin/src/lib/api/types/profile.ts
retini: Done in Rename [`consent to consents in ProfileEventContext`](https://github.com/krenalis/krenalis/pull/2355/changes/136fa4cb76db04029dab55f9652a68f43d5b6006)

admin/src/components/routes/Privacy/Privacy.tsx
retini: Done in [`Improve the warning shown when changing a purpose code`](https://github.com/krenalis/krenalis/pull/2355/changes/1f2793b7e837b85fa30aab42f0a73f834662122c)

admin/src/components/routes/PipelineWrapper/PipelineConsents.tsx
retini: Done in [`Make the privacy sentence toggle the required consents checkbox`](https://github.com/krenalis/krenalis/pull/2355/changes/148c7bc470bf7256993b3120839e8d5ecf38c00c)

gazerro (approved):

PR #2360, opened 2026-07-08
cmd, core: add `Request-Id` and `Sync-Token` headers to APIs

### LINKED ISSUE #193: Investigate on what the APIs should guarantee
In this scenario:

1. an entity is created through the APIs
2. such entity is referred in a following call

If the state updates, consequent to the first call, are not propagated to the current instance state ­— and to the other instances — the second call may not see the updates.

For example, if an API call creates a connection and a following call tries to add an action to it, it may happen that Krenalis returns "Not found" as the new connection, written to the database through the transaction, has still not been added to the state.

We should discuss if this behavior may be expected or not, and more generally, what should and should not be guaranteed by our APIs.

gazerro:
```
cmd, core: add `Request-Id` and `Sync-Token` headers to APIs

Every /v1 API response now includes a Request-Id header containing a
unique identifier for the request. Successful 200 OK responses also
include a Sync-Token header. Clients can send this token in a subsequent
request to ensure that it observes the same state as the response that
produced the token, or a newer one.

Request-Id is included in normal API responses and API error responses
after the request enters the /v1 API server. It is also added to
request-scoped logs, making it easier to trace errors back to the API
request that caused them.

Sync-Token is included only in successful 200 OK responses.

Closes #193
```


gazerro:
See also https://github.com/krenalis/krenalis.com/pull/44.

zapateo (changes requested):
As discussed verbally, what is currently a single header should be split into two, serving two different purposes:

* A header containing the request ID, which allows the request to be associated with any logs in the event of an error.
* A header containing the state version, which in turn can be passed back to request a minimum state version after which the request should be executed.

cmd/internal/synctoken/codec.go
```diff
@@ -0,0 +1,104 @@
...
+	token = c.aead.Seal(token, nonce, plaintext[:], []byte(associatedData))
+	return base58.EncodeToString(token), nil
+}
+
+// Decode decodes a Sync-Token value and returns the state version.
+//
+// It returns errInvalidSyncToken if value is malformed, has an unexpected
+// size, was generated with another key, or fails authentication.
+func (c *Codec) Decode(value string) (int, error) {
+	token, err := base58.DecodeString(value)
```
zapateo, line 92: Before decoding, we should check the length; if it's longer than 50 characters, an error should be returned.

Otherwise, given the complexity (quadratic?) of Base58 decoding, an attacker (even unauthenticated) could consume the CPU for several minutes with a single request.

cmd/internal/synctoken/responsewriter.go
```diff
@@ -0,0 +1,82 @@
...
+	return &ResponseWriter{
+		ResponseWriter: rw,
+		codec:          codec,
+		nonce:          nonce,
+		stateVersion:   stateVersion,
+	}
+}
+
+// Unwrap returns the wrapped response writer.
+func (w *ResponseWriter) Unwrap() http.ResponseWriter {
```
zapateo, line 41: This method is unused, consider if it's worth keeping it or not.

cmd/apis_server.go
```diff
@@ -155,14 +162,30 @@ func newAPIsServer(core *core.Core, runsOnHTTPS bool, javaScriptSDKURL, external
...
 		return
 	}
 
+	// Generate a random request ID to use as both the Request-Id value
+	// and the nonce for the Sync-Token.
+	var rawRequestID [synctoken.NonceSize]byte
+	_, _ = rand.Read(rawRequestID[:])
+
+	// Encode the request ID as Base58 for use in the Request-Id header and
+	// puts it in the request's context.
```
zapateo: This should be `put` instead of `puts`, shouldn't it?

core/internal/initdb/upgrade.go
zapateo, line 1: An upgraded database (on the left) is different from a database recreated on this branch from scratch (on the right):

Image

core/internal/state/load.go
```diff
@@ -631,15 +631,16 @@ func (state *State) load(ctx context.Context, oauthCredentials map[string]*OAuth
 		return fmt.Errorf("cannot load primary sources: %s", err)
 	}
 
-	// Read the last notification ID.
-	var latest int64
-	err = tx.QueryRow(ctx, "SELECT COALESCE(MAX(id),0) FROM notifications").Scan(&latest)
+	// Read the last notification version.
+	var version int
```
zapateo, line 635: There's something I'm missing here, maybe it's not important, but I'd like to clarify:

* Why was the variable type changed from `int64` to `int`? Is it just a matter of code cleanliness, or is there more to it?

* Below, the comparison still occurs with the constant `math.MaxInt64`. Is this intentional?

zapateo (changes requested):

cmd/internal/synctoken/codec.go
gazerro: Fixed in commit [Check the maximum token length in Codec.Decode](https://github.com/krenalis/krenalis/pull/2360/changes/e6dbb8fc7d7dffcb453039b1e691338cc79b4ed3).

cmd/internal/synctoken/responsewriter.go
gazerro: See https://pkg.go.dev/net/http#NewResponseController.

In practice, this exposes the underlying writer so that the `net/http` package can access it when needed. This provides an additional layer of compatibility, alongside embedding the original writer.

I am not entirely sure whether this will be useful in our specific case, but it comes at virtually no cost.

cmd/apis_server.go
gazerro: Fixed in commit [Fix comment](https://github.com/krenalis/krenalis/pull/2360/changes/e6281c519b0596352924342704a1e24bf7a36154).

core/internal/state/load.go
gazerro: The Krenalis codebase assumes that the software always runs on a 64-bit platform. For this reason, `int` is used instead of `int64`, since `int` is always 64 bits in the supported environments.

core/internal/initdb/upgrade.go
gazerro: Fixed in commit [Rename upgraded NOT NULL constraints](https://github.com/krenalis/krenalis/pull/2360/changes/a460ebd796acfb72350355ed00fbc53d23b87926).

cmd/apis_server.go
zapateo, line 1: _(I'm noting this here because I don't know where else to put it.)_

If an error occurs while Krenalis is receiving an event, the request ID isn't logged.

Furthermore, this specific case I'm mentioning here in the example (invalid GZIP content) causes an Internal Server Error when it shouldn't. For now, we're fortunate that this is the case, so we can test the request ID issue. 😂

So, if I'm correct, we should:

1. Fix the Request ID log issue in this PR
2. Fix the gzip issue in a separate PR

## How to test this

Run:

```bash
printf 'this-is-not-gzip' | curl -i \
  -H 'Content-Type: text/plain' \
  -H 'Content-Encoding: gzip' \
  --data-binary @- \
  http://localhost:2022/v1/events
```

You should get something like this as response from Krenalis:

```
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Origin: *
Content-Type: text/plain; charset=utf-8
Request-Id: 1amPuPTigiw5HSbC
X-Content-Type-Options: nosniff
Date: Wed, 15 Jul 2026 14:06:09 GMT
Content-Length: 22

Internal Server Error
```

While the Krenalis log contains just this, without the request ID:

```
ERROR core/events/collector: an error occurred creating gzip reader error="gzip: invalid header"
```

zapateo (changes requested):

cmd/apis_server.go
gazerro: > 1. Fix the Request ID log issue in this PR

Done in commits [Make request IDs accessible across core packages](https://github.com/krenalis/krenalis/pull/2360/changes/eaf85e4f7250bdf39706082ef0b74ab2a3857203) and [Link request ID helpers in Core docs](https://github.com/krenalis/krenalis/pull/2360/changes/87c759f78bc8f1b3785505cadf8f6ced6130aa73).

> 2. Fix the gzip issue in a separate PR

Done in PR #2371.

zapateo (approved):

PR #2366, opened 2026-07-10
cmd, core: add metrics grouped by workspace, connection, and pipeline

gazerro:
```
cmd, core: add metrics grouped by workspace, connection, and pipeline

Pipeline metrics can now be grouped by workspace, connection, or
pipeline. Each request must specify exactly one non-empty grouping
parameter.

Unknown identifiers and identifiers outside the request scope are
ignored. The response includes one series for each selected group with
metric data in the requested time range.

Pipeline metrics no longer depend on a live pipeline row, so historical
data remains available after a pipeline is deleted.

This allows the Admin console to query connection-level metrics without
requiring every historical pipeline to still exist.
```


gazerro:
See documentation: https://github.com/krenalis/krenalis.com/pull/47

core/internal/metrics/metrics.go
zapateo, line 1: The documentation for the various metric endpoints on krenalis.com doesn't specify the resolution of each time bucket returned by the endpoints.

I think it would be helpful to specify this in the method description, to clearly understand what the method does (because the resolution greatly changes the possible applications and uses of that method), and also to specify it in the "passed" and "failed" fields. Maybe it's just a matter of adding a small note on the resolution, nothing long, just a few words.

So, for example, regarding the **Get metrics per day** method, I would describe it with something like:

> Retrieves pipeline metrics aggregated by day for the specified number of days up to the current time.

While in the "passed" and "failed" fields, in my opinion something like this is sufficient:

> Each item represents one daily time bucket.

What do you think?

core/pipeline.go
```diff
@@ -144,7 +144,8 @@ const (
 type Target int
 
 const (
-	TargetEvent Target = iota + 1
```
zapateo, line 147: Here, the `TargetNone` is introduced, but the `String` method underneath hasn't been modified.

I don't know if this is intentional or not, but by introducing the `TargetNone` constant, it effectively becomes a valid value of `Target`, so I don't expect calling the `String` method on it to cause a panic.

What do you think?

core/internal/metrics/pipelines.go
```diff
@@ -32,10 +33,48 @@ type Error struct {
...
+	Connections []string
+	Pipelines   []string
+	Target      Target
+}
+
+// Target represents a target.
+type Target int
+
+const (
+	TargetNone = iota
```
zapateo: See the other comment: https://github.com/krenalis/krenalis/pull/2366/changes#r3593573538.

core/pipeline.go
```diff
@@ -144,7 +144,8 @@ const (
 type Target int
```
zapateo, line 144: Here, the `Target` type is defined, but then the constants don't have that type, and are implicitly `int`. They should look something like this:

```go
const (
	TargetNone Target = iota
	TargetEvent
	TargetUser
	TargetGroup
)
```

core/internal/metrics/pipelines.go
```diff
@@ -32,10 +33,48 @@ type Error struct {
...
+// Selection describes which metric series are returned.
+type Selection struct {
+	Workspaces  []string
+	Connections []string
+	Pipelines   []string
+	Target      Target
+}
+
+// Target represents a target.
+type Target int
```
zapateo, line 59: See the other comment: https://github.com/krenalis/krenalis/pull/2366/changes#r3593693243.

core/organization.go
```diff
@@ -1375,6 +1583,59 @@ func validateMemberEmail(email string) error {
 	return nil
 }
 
+// validateMetricsSelection validates a pipeline metrics selection.
+func validateMetricsSelection(selection MetricSelection) error {
```
zapateo: This function doesn't impose a limit on the number of pipeline/connection/workspace IDs passed.

In practice, even if one passed thousands of random IDs, they would be filtered out by `filterMetricsSelection` (at least those that don't exist as identifiers in the request scope), but in general, I'm thinking if it's worth introducing a cap on the number of IDs regardless.

What do you think?

core/internal/metrics/metrics.go
```diff
@@ -126,7 +127,12 @@ func (c *Collector) Failed(step Step, pipeline string, count int, message string
 	c.mu.Lock()
 	m, ok := c.metrics[pipeline]
 	if !ok {
-		m = &metrics{}
+		var err error
+		m, err = c.newMetrics(pipeline)
+		if err != nil {
```
zapateo, line 132: Here, if an error is returned from `newMetrics`, the error is discarded and not logged.

In the current implementation, `newMetrics` can only return an error if the pipeline is not found, and so that's fine. However, if it also returns other types of errors in the future, we risk discarding significant errors.

One thing we could do is change `newMetrics` to document the fact that it only returns an error if the pipeline doesn't exist, or make it return a boolean, making it even more explicit.

Tell me what you think and see if it's worth it.

core/internal/metrics/metrics.go
```diff
@@ -203,7 +209,12 @@ func (c *Collector) Passed(step Step, pipeline string, count int) {
...
```
zapateo, line 215: See the other comment: https://github.com/krenalis/krenalis/pull/2366/changes#r3593942348.

core/organization.go
```diff
@@ -826,6 +827,184 @@ func (this *Organization) Members(ctx context.Context) ([]*Member, error) {
...
+
+	// Validate selection.
+	err := validateMetricsSelection(selection)
+	if err != nil {
+		return Metrics{}, err
+	}
+
+	// Validate workspace.
+	if workspace != "" && !IsValidID(workspace) {
+		return Metrics{}, errors.BadRequest("workspace %s is not valid", workspace)
```
zapateo: We generally use the verb `%q` for identifiers, or am I wrong?

admin/src/components/routes/ConnectionMetrics/ConnectionMetrics.tsx
zapateo, line 1: Consider this scenario:

1. I have two pipelines importing users from the same connection
2. I delete one of the two pipelines
3. I want to view metrics only for the remaining pipeline

Currently, with this change, the pipeline selection menu disappears when there is only one pipeline, so the data I see is always added to that of the removed pipelines.

We should therefore consider keeping the pipeline selection menu, even when there is only one.

zapateo, line 1: In the **Error log** section of the pipeline metrics screen, only existing pipeline errors are shown, while the count displayed above also includes deleted pipelines.

One thing we could consider, if it's not complicated, is to display a small message at the bottom explaining why the count appears inconsistent if the failed count doesn't match the error count shown below.

zapateo (changes requested):

admin/src/components/routes/ConnectionMetrics/ConnectionMetrics.tsx
gazerro: @retini

gazerro: @retini

core/pipeline.go
gazerro: It was intentional, but you're right. I've fixed it in [Fix Target type constants and methods](https://github.com/krenalis/krenalis/pull/2366/changes/e031cd85cb0d3834007774786486f9b613b69ad9).

core/internal/metrics/pipelines.go
gazerro: See [Fix Target type constants and methods](https://github.com/krenalis/krenalis/pull/2366/changes/e031cd85cb0d3834007774786486f9b613b69ad9).

core/pipeline.go
gazerro: See [Fix Target type constants and methods](https://github.com/krenalis/krenalis/pull/2366/changes/e031cd85cb0d3834007774786486f9b613b69ad9).

core/internal/metrics/pipelines.go
gazerro: See [Fix Target type constants and methods](https://github.com/krenalis/krenalis/pull/2366/changes/e031cd85cb0d3834007774786486f9b613b69ad9).

retini (commented):
@gazerro @zapateo As discussed, rather than reviewing the existing admin code, I’m going to update it directly and take ownership of the requested changes.

I refactored the code in the commit [`Review and update the admin code`](https://github.com/krenalis/krenalis/pull/2366/changes/372fa16bd43529733322a1b2bf5dc0bf83555178), where I also updated the Metrics page so that the **Events** tab is enabled only when the connection supports events and remains disabled otherwise.

I’ll now complete the requested changes and then request a review from @zapateo.

admin/src/components/routes/ConnectionMetrics/ConnectionMetrics.tsx
retini: Done in [`Keep pipeline selector visible even with a single pipeline`](https://github.com/krenalis/krenalis/pull/2366/changes/b2480f85f7fdc2029f70cdf57c7c286f14ebd72a)

core/organization.go
gazerro: Fixed in commit [Limit workspace, connection, and pipeline selections to 100 items](https://github.com/krenalis/krenalis/pull/2366/changes/f74bf8f0c1ec717bfc1e1714aa3d2f6833a303d6) and documented in commit [Document metrics limits](https://github.com/krenalis/krenalis.com/pull/47/changes/9225b4f6ed345be1e315af159bc065f29ea08e2a).

admin/src/components/routes/ConnectionMetrics/ConnectionMetrics.tsx
retini: Done in [`Note hidden failures from deleted pipelines in error table`](https://github.com/krenalis/krenalis/pull/2366/changes/41fedb56c7024f18978ca10c8f0585dd54481ff7)

core/internal/metrics/metrics.go
gazerro: Done in commit [Clarify when newMetrics returns an error](https://github.com/krenalis/krenalis/pull/2366/changes/e4cbdf991706d820638d8e5614629a9aebf971a1).

core/internal/metrics/metrics.go
gazerro: Done in commit [Clarify when newMetrics returns an error](https://github.com/krenalis/krenalis/pull/2366/changes/e4cbdf991706d820638d8e5614629a9aebf971a1).

core/organization.go
gazerro: We use `%s` for identifiers, but this value is not an identifier, so `%q` should be used instead. Fixed in commit [Fix error message](https://github.com/krenalis/krenalis/pull/2366/changes/426fb722221525013a69c7d6d1478cede7806916).

core/internal/metrics/metrics.go
gazerro: Fixed in commit [Fix documentation](https://github.com/krenalis/krenalis.com/pull/47/changes/614dca1ea8f6c888df4e726a4ec111b26217d465).

cmd/endpoints.go
```diff
@@ -67,10 +67,10 @@ func endpoints(s *apisServer) map[string]endpointHandler {
 		"GET    /organizations/{id}":                          api.Organization,                     /* Needs organizations API key */
 		"GET    /organizations":                               api.Organizations,                    /* Needs organizations API key */
 		"GET    /pipelines/errors/{start}/{end}":              workspace.PipelineErrors,
-		"GET    /pipelines/metrics/dates/{start}/{end}":       workspace.PipelineMetricsPerDate,
-		"GET    /pipelines/metrics/days/{days}":               workspace.PipelineMetricsPerDay,
-		"GET    /pipelines/metrics/hours/{hours}":             workspace.PipelineMetricsPerHour,
-		"GET    /pipelines/metrics/minutes/{minutes}":         workspace.PipelineMetricsPerMinute,
+		"GET    /pipelines/metrics/dates/{start}/{end}":       organization.PipelineMetricsPerDate,
```
zapateo, line 70: This endpoint allows the request of metrics for a wide range of days.

For example, consider this API call via `curl`, where I request metrics for two pipelines that import users:

```bash
curl -sS \
  -H "Authorization: Bearer api_vwyauuZl5NLKf8wE2AHha1dbgNYETcmc0CHaEN1KdKHCsinHm" \
  "http://127.0.0.1:2022/v1/pipelines/metrics/dates/1970-01-01/2262-04-11?pipelines=dTEycAfEFHMf,nEYj7Y8NteSG"
```

Krenalis accepts the call and prepares the response, despite the day range being approximately 292 years (~106000 days).

Shouldn't this be limited during validation? Other metric methods accept a limited range.

As a consequence (but this needs to be eventually investigated), may this code then allocate too much memory?

https://github.com/krenalis/krenalis/blob/e031cd85cb0d3834007774786486f9b613b69ad9/core/internal/metrics/pipelines.go#L225-L256

core/internal/metrics/metrics.go
zapateo, line 1: In this file, in the `aggregate` function, there is some behavior that should be investigated, as discussed verbally.

zapateo (changes requested):

cmd/endpoints.go
gazerro: You've identified a real issue.

We could limit the date range to 731 days. That would cover any two consecutive calendar years, including a leap year, while still allowing users to compare metrics across a two-year period.

The current limit of 100 pipelines, 100 connections, or 100 workspaces does not make the query cost predictable. For 100 pipelines over 731 days, the database would read only slightly more than 100 × 731 consolidated daily rows. However, 100 connections or 100 workspaces may include many more pipelines, so the actual number of rows read could be much higher.

We could count the underlying pipelines before running the metrics query and reject requests above a certain threshold. The downside is that callers would not know in advance whether their request is valid. For example, a request for a single workspace over an allowed date range could still fail simply because that workspace contains too many pipelines. That would make the API limits hard to understand and the result of a request difficult to predict.

A better alternative may be to consolidate metrics at the connection and workspace levels as well. Since pipelines cannot move between connections, and connections cannot move between workspaces, maintaining these additional rollups should be relatively straightforward and would not create ambiguity in historical data.

For example, assuming:

* 1,000 pipelines;
* 100 connections;
* 10 workspaces;

the number of stored metric series would increase from 1,000 to 1,110. That is only an 11% increase, not three times as much.

This would make query costs predictable and allow us to expose simple limits: a maximum range of 731 days and a maximum of 100 selected entities, regardless of whether the requested level is pipelines, connections, or workspaces.

Alternatively, or in addition, we could limit the product of the number of selected entities and the number of requested days.

core/internal/metrics/metrics.go
gazerro: This was fixed by the commit [core/internal/metrics: fix daily aggregation interval](https://github.com/krenalis/krenalis/commit/1acf5303c00cb1d5084adf3410d8ceb5004e20b7), correct?

zapateo: Yes!

cmd/endpoints.go
zapateo: As discussed verbally, the changes to be made are therefore:

* Increase the number of requestable entities from 100 to 1,000 (I'm referring to the values ​​encoded in `validateMetricsSelection`)
* Ensure that the product of the number of days requested and the number of requested entities is <= 50,000.

The number of days is therefore not limited per se.

We'll postpone internal optimization, i.e., consolidating the metrics at the connection and workspace levels, until later.

core/internal/metrics/metrics.go
```diff
@@ -439,8 +465,8 @@ func (c *Collector) store(timeslot int32, metrics map[string]*metrics) {
 	if i > 0 {
 
 		c.buf.WriteString("\n) INSERT INTO pipelines_metrics AS m " +
-			`(pipeline, timeslot, passed_0, passed_1, passed_2, passed_3, passed_4, passed_5, failed_0, failed_1, failed_2, failed_3, failed_4, failed_5)` +
-			` SELECT t.* FROM t WHERE EXISTS (SELECT 1 FROM pipelines p WHERE p.id = t.pipeline)` +
+			`(organization, workspace, connection, pipeline, target, timeslot, passed_0, passed_1, passed_2, passed_3, passed_4, passed_5, failed_0, failed_1, failed_2, failed_3, failed_4, failed_5)` +
```
zapateo, line 468: I have a doubt about this query.

If the organization is deleted just before running this query, the query would fail because there is a constraint on `organization`:

https://github.com/krenalis/krenalis/blob/d0ac10bd2b86d208db10f702e14d81440ee346cb/core/internal/initdb/schema.sql#L243

The consequences would be:

* The query would be invalid, and the information it was supposed to add would be lost; this information refers not only to the deleted organization, but also to other metrics of organizations that still exist.
* The backoff does not impose any limit on the number of retries, so it would continue to [try for 15 minutes](https://github.com/krenalis/krenalis/blob/fd0f0999268c54b8d90a89138de7de6ff06b6572/tools/backoff/backoff.go#L55-L56) to execute a query that would always fail and would block the goroutine that writes the metrics.

On `main` we had a pipeline guard, shouldn't there be one on the organization here instead? That is, should we insert if the organization actually exists?

Does this make sense to you?

core/organization.go
```diff
@@ -1375,6 +1593,68 @@ func validateMemberEmail(email string) error {
 	return nil
 }
 
+// validateMetricsSelection validates a pipeline metrics selection.
+func validateMetricsSelection(selection MetricSelection) error {
```
zapateo: `validateMetricsSelection` should be moved after `validateMemberToSet`.

cmd/api_test.go
```diff
@@ -78,3 +82,106 @@ func TestSplitQueryParameters(t *testing.T) {
 		})
 	}
 }
+
+// TestParsePipelineMetricsSelectionConnectionTarget verifies connection and
+// target query parameters for pipeline metrics requests.
+func TestParsePipelineMetricsSelectionConnectionTarget(t *testing.T) {
```
zapateo: This function, as well as `TestParsePipelineMetricsSelectionWorkspaces`, test the function `parseMetricsSelection` which is not defined in `api.go`, but in `organization.go`.

Is this intended? Or should these tests be placed, for example, in `organization_test.go`?

admin/src/components/routes/ConnectionMetrics/ConnectionMetrics.tsx
```diff
@@ -478,6 +504,12 @@ const ConnectionMetrics = () => {
 		}
 	}
 
+	const funnelData = isUsersSelected ? userFunnelData : eventFunnelData;
+	const errors = isUsersSelected ? userPipelinesErrors : eventPipelinesErrors;
+	const totalFailed = computeTotalFailed(funnelData);
+	const totalErrorsCount = errors.reduce((sum, e) => sum + e.count, 0);
+	const hasHiddenFailures = totalFailed > totalErrorsCount;
```
zapateo: I think we have a problem with this comparison.

`totalErrorsCount` is based on a Krenalis API call with a limit of 50 errors, if I'm not mistaken:

https://github.com/krenalis/krenalis/blob/41fedb56c7024f18978ca10c8f0585dd54481ff7/admin/src/components/routes/ConnectionMetrics/ConnectionMetrics.tsx#L375

As a result, in all cases where the errors are more than 50, `hasHiddenFailures` would have a wrong value (a false positive) and then this message would be displayed:

> Some of the failures counted above are not listed here because they occurred in pipelines that have since been deleted.

when in reality, perhaps no pipeline was ever deleted.

We should discuss about this.

zapateo (changes requested):

core/organization.go
gazerro: Fixed in commit [Reorder methods in core/organization.go](https://github.com/krenalis/krenalis/pull/2366/changes/11ed0f178bafe8f8b16b7cd22ab3ab0f0cf49bab).

cmd/api_test.go
gazerro: Fixed in commit [Move tests from api_test.go to organization_test.go](https://github.com/krenalis/krenalis/pull/2366/changes/54a98ab34abdc44e64f9affd0b8dc49fd8b4d8b5).

See also commit [Rename core's organization_metrics_test.go to organization_test.go](https://github.com/krenalis/krenalis/pull/2366/changes/d61f2d9c3d55f1edb3037786efc7ea0bd2b15c51).

core/internal/metrics/metrics.go
gazerro: Fixed in commit [Fix metrics storage for deleted organizations](https://github.com/krenalis/krenalis/pull/2366/changes/77d07d556e3428c60dea12f5a5d54a78c0479745).

cmd/endpoints.go
gazerro: Changed in comits [Increase the number of requestable entities from 100 to 1,000](https://github.com/krenalis/krenalis/pull/2366/changes/4c7109ac245e009a31ca88da2506d3b8a27dbc91) and [Limit metric requests to 60,000 data points](https://github.com/krenalis/krenalis/pull/2366/changes/f52293507003a8c1f185af0ae0277d0dace38fe6).

admin/src/components/routes/ConnectionMetrics/ConnectionMetrics.tsx
zapateo: @retini, I'm assigning this conversation to you.

Considering:

1. The issue I mentioned above;
2. The fact that the Admin currently has no information on how many total errors there are, since it always displays a fraction (max 50) and doesn't know the total (Krenalis doesn't return the total number of errors, and to do so would require a specific modification that isn't worth it now);

Perhaps the simplest thing is to ensure there's always an **ⓘ** that shows a more generic message like:

> These are the last 50 pipeline errors for this connection. They don't include errors from any deleted pipelines.

Decide which format you think is best.

retini: Done in c8fe5ff3a44989b073244b4e38ef91fcca45d557

cmd/endpoints.go
zapateo: Regarding terminology, as discussed in person, I'd favor eliminating the term "data points" and simply avoiding naming the product entities * days.

Furthermore, if I understand correctly, the term "items" is sometimes used in place of "entries"; in this regard, I'd favor using the latter.

In essence, I'd propose a change like this; see if it makes sense to you:

```diff
diff --git a/core/organization.go b/core/organization.go
index 216a60fbd..73eebbf73 100644
--- a/core/organization.go
+++ b/core/organization.go
@@ -872,7 +872,7 @@ const (
 	Day    = MetricUnit(metrics.Day)
 )
 
-const maxMetricDataPoints = 60_000 // maximum number of data points in a metrics response
+const maxEntryDays = 60_000 // maximum product between the number of entries in the selection and the number of days in the requested date range
 
 // PipelineMetricsPerDate returns metrics aggregated by day for the time
 // interval between the specified start and end dates.
@@ -884,11 +884,10 @@ const maxMetricDataPoints = 60_000 // maximum number of data points in a metrics
 // day of the end date.
 //
 // Exactly one of workspaces, connections, and pipelines in selection must be
-// non-nil, with between 1 and 1,000 items.
+// non-nil, with between 1 and 1,000 entries.
 //
-// The requested metrics cannot exceed the maximum of 60,000 data points,
-// computed as the number of days in the date range multiplied by the number
-// of entries in the selection.
+// The number of entries in the selection multiplied by the number of days in
+// the date range cannot exceed 60,000.
 func (this *Organization) PipelineMetricsPerDate(ctx context.Context, start, end time.Time, workspace string, selection MetricSelection) (Metrics, error) {
 
 	this.core.mustBeOpen()
@@ -913,8 +912,8 @@ func (this *Organization) PipelineMetricsPerDate(ctx context.Context, start, end
 	if err != nil {
 		return Metrics{}, err
 	}
-	if days*entries > maxMetricDataPoints {
-		return Metrics{}, errors.BadRequest("requested metrics exceed the maximum of 60,000 data points (%d days × %d selection entries)", days, entries)
+	if days*entries > maxEntryDays {
+		return Metrics{}, errors.BadRequest("requested metrics exceed the maximum: %d selection entries × %d days is more than 60,000", entries, days)
 	}
 
 	// Validate workspace.
@@ -960,7 +959,7 @@ func (this *Organization) PipelineMetricsPerDate(ctx context.Context, start, end
 // and [1,30] for days.
 //
 // Exactly one of workspaces, connections, and pipelines in selection must be
-// non-nil, with between 1 and 1000 items.
+// non-nil, with between 1 and 1,000 entries.
 func (this *Organization) PipelineMetricsPerTimeUnit(ctx context.Context, number int, unit MetricUnit, workspace string, selection MetricSelection) (Metrics, error) {
 
 	this.core.mustBeOpen()
diff --git a/core/organization_test.go b/core/organization_test.go
index 066a826b2..0f0554da4 100644
--- a/core/organization_test.go
+++ b/core/organization_test.go
@@ -11,9 +11,10 @@ import (
 	"time"
 )
 
-// TestPipelineMetricsPerDateRejectsTooManyDataPoints verifies that requests
-// exceeding the maximum number of data points are rejected.
-func TestPipelineMetricsPerDateRejectsTooManyDataPoints(t *testing.T) {
+// TestPipelineMetricsPerDateRejectsTooManyEntryDays verifies that requests
+// exceeding the maximum product between the number of entries in the selection
+// and the number of days in the date range are rejected.
+func TestPipelineMetricsPerDateRejectsTooManyEntryDays(t *testing.T) {
 	organization := Organization{core: &Core{}}
 	start := time.Date(2026, time.January, 1, 0, 0, 0, 0, time.UTC)
 	tests := []struct {
@@ -33,8 +34,8 @@ func TestPipelineMetricsPerDateRejectsTooManyDataPoints(t *testing.T) {
 			if err == nil {
 				t.Fatal("expected error, got nil")
 			}
-			if !strings.Contains(err.Error(), "requested metrics exceed the maximum of 60,000 data points") {
-				t.Fatalf("expected data point limit error, got %v", err)
+			if !strings.Contains(err.Error(), "requested metrics exceed the maximum") {
+				t.Fatalf("expected entry-day limit error, got %v", err)
 			}
 		})
 	}

```

zapateo (changes requested):

cmd/endpoints.go
gazerro: See commit [Replace "data point" with "entry-day"](https://github.com/krenalis/krenalis/pull/2366/changes/2228f16d61184cbb2be69ab7ee3331104de72174).

zapateo (approved):

PR #2373, opened 2026-07-20
core/internal/collector: fix API key scoping for event ingestion

gazerro:
```
core/internal/collector: fix API key scoping for event ingestion

Fixes four issues affecting event ingestion when using an API key that
is not restricted to a specific workspace:

a) The collector overwrites the AccessKey.Workspace field from the state
   package, even though state fields must never be modified.
   This introduces a data race and can cause an API key that was
   originally not restricted to a workspace to become unexpectedly bound
   to the workspace specified in the request through the
   Krenalis-Workspace header.

b) The collector does not verify that the workspace specified in the
   Krenalis-Workspace header belongs to the organization associated with
   the API key. As a result, anyone who knows a workspace ID can use an
   API key from one organization to ingest events into a workspace owned
   by another organization.

c) The collector does not verify that the connection specified in the
   request body belongs to the organization associated with the API key.
   As a result, anyone who knows a connection ID can use an API key from
   one organization to ingest events into a connection owned by another
   organization.

d) When an organization is disabled, no information about its resources
   should be disclosed. However, the collector reveals whether the
   workspace specified in the Krenalis-Workspace header exists.
   Combined with the issue described in point b), this information is
   also disclosed when the workspace belongs to a different
   organization.
```

### DIFF
--- a/core/internal/collector/collector.go
+++ b/core/internal/collector/collector.go
@@ -707,7 +707,7 @@ func (c *Collector) serveEvents(w http.ResponseWriter, r *http.Request) error {
 			}
 			if ws == nil {
 				connection, ok = c.state.Connection(id)
-				if ok && connection.Organization() != org {
+				if ok && connection.Organization().ID != org.ID {
 					connection = nil
 				}
 			} else {
--- a/core/internal/collector/collector.go
+++ b/core/internal/collector/collector.go
@@ -665,6 +665,14 @@ func (c *Collector) serveEvents(w http.ResponseWriter, r *http.Request) error {
 			if key.Type != state.AccessKeyTypeAPI {
 				return errors.Unauthorized("API key in the Authorization header is invalid")
 			}
+			org, ok := c.state.Organization(key.Organization)
+			if !ok {
+				return errors.Unauthorized("API key in the Authorization header is invalid")
+			}
+			if !org.Enabled {
+				return errors.Unprocessable("OrganizationDisabled", "organization %s is disabled", org.ID)
+			}
+			var ws *state.Workspace
 			if header, ok := r.Header["Krenalis-Workspace"]; ok {
 				if len(header) > 1 {
 					return errors.BadRequest(`request contains multiple "Krenalis-Workspace" headers`)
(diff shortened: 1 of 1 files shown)


core/internal/collector/collector.go
```diff
@@ -694,21 +705,17 @@ func (c *Collector) serveEvents(w http.ResponseWriter, r *http.Request) error {
...
 			}
-			if key.Workspace == "" {
-				connection, _ = c.state.Connection(id)
-			} else {
-				workspace, ok := c.state.Workspace(key.Workspace)
-				if !ok {
-					return errors.Unauthorized("API key in the Authorization header is invalid")
+			if ws == nil {
+				connection, ok = c.state.Connection(id)
+				if ok && connection.Organization() != org {
```
zapateo: Here, the comparison between organizations is done by comparing pointers.

What happens if an organization is updated in the meantime (by calling `replaceOrganization`) in the state? Am I wrong, or is there a case where a comparison that should be positive actually turns out to be negative, because the organization is logically the same but the pointers are different?

Shouldn't the IDs be compared, or am I missing something?

core/internal/collector/collector.go
zapateo, line 1: I would also document in the commit message the problem we have on `main` regarding connections and which is solved with this branch, namely the fact that the ingestion of an event that specifies a connection but does not use a workspace-restricted key nor pass the `Krenalis-Workspace` header can ingest into another organization (since it was not checked that the connection belonged to the key's org, while now it does).

zapateo (changes requested):

core/internal/collector/collector.go
gazerro: I completely rewrote the commit message to make it clearer and added a point about connection scoping.

core/internal/collector/collector.go
gazerro: Fixed in commit [Compare organizations by ID instead of pointer identity](https://github.com/krenalis/krenalis/pull/2373/changes/e59c3943dff8dba25d4a7859bb019c5962a6c477).

zapateo (approved):

PR #2376, opened 2026-07-21
cmd: reject conflicting workspace authentication

### LINKED ISSUE #2098: Handle the case where two different workspaces are specified in API calls
This method:

https://github.com/krenalis/krenalis/blob/6ef39a1d47fb8ebcb7c905594f8a02ebe02220e3/cmd/apis_server.go#L250

it doesn't seem to handle the case in which the workspace is passed simultaneously both via the header and implicitly via an API key that operates on a single workspace.

What happens if the two workspaces, passed differently, are different from each other? One possibility would be to return error.

Additionally, whatever we decide to do, we should explicitly document it on krenalis.com.

gazerro:
```
cmd: reject conflicting workspace authentication

Return a 400 Bad Request error when a workspace-scoped key is used
together with the Krenalis-Workspace header. This brings the behavior in
line with the collector.

Also return a key-related error when the organization or workspace
associated with the key no longer exists, so the key is treated as
invalid or nonexistent.

Closes #2098
```

### DIFF
--- a/cmd/apis_server.go
+++ b/cmd/apis_server.go
@@ -374,36 +374,44 @@ func (s *apisServer) authenticateRequest(r *http.Request) (*core.Organization, *
 		}
 		org, err := s.core.Organization(organizationID)
 		if err != nil {
+			if _, ok := err.(*errors.NotFoundError); ok {
+				err = errors.Unauthorized("API key in the Authorization header of the request does not exist")
+			}
 			return nil, nil, err
 		}
 		if !org.Enabled {
 			return nil, nil, errors.Unprocessable(core.OrganizationDisabled, "organization %s is disabled", org.ID)
 		}
-		// If the key is restricted to a workspace, return the workspace as well.
-		if workspaceID != "" {
-			ws, err := org.Workspace(workspaceID)
+		// If the Krenalis-Workspace header is present, return the workspace as well.
+		if header, ok := r.Header["Krenalis-Workspace"]; ok {
+			if len(header) > 1 {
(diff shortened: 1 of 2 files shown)


cmd/apis_server.go
```diff
@@ -374,36 +374,44 @@ func (s *apisServer) authenticateRequest(r *http.Request) (*core.Organization, *
...
 		if !org.Enabled {
 			return nil, nil, errors.Unprocessable(core.OrganizationDisabled, "organization %s is disabled", org.ID)
 		}
+		// If the Krenalis-Workspace header is present, return the workspace as well.
+		if header, ok := r.Header["Krenalis-Workspace"]; ok {
+			if len(header) > 1 {
+				return nil, nil, errors.BadRequest("request contains multiple Krenalis-Workspace headers")
+			}
+			if workspaceID != "" {
+				return nil, nil, errors.BadRequest(`"Krenalis-Workspace" header cannot be provided with a workspace restricted key`)
```
zapateo, line 391: The commit message states:

> Return an **authentication error** when a workspace-scoped key is used
together with the Krenalis-Workspace header.

but the code returns a 400 Bad Request.

It's just an inaccuracy in the commit message, right? I assume the code is correct, that this is indeed a Bad Request (as [the collector already does](https://github.com/krenalis/krenalis/blob/8495eec37281d89c39da37882186fd155cce9bb4/core/internal/collector/collector.go#L680-L682)).

zapateo (changes requested):

cmd/apis_server.go
gazerro: Fixed.

zapateo (approved):

PR #2382, opened 2026-07-23
core, cmd: add API rate limiting

gazerro:
```
core, cmd: add API rate limiting (#2382)

Add configurable per-minute refill rates and maximum capacities for the
global budget for platform management API requests, organization-level
requests, workspace requests, and event ingestion. Enforce the
appropriate budget for HTTP API requests, MCP tool calls, and event
ingestion.

Implement a distributed rate limiter backed by authoritative PostgreSQL
buckets and batched capacity leasing. Each process consumes leased
capacity locally and coalesces refill requests, avoiding a database
query for every operation. Start local buckets with small lease targets
and increase those targets under sustained demand, reducing capacity
fragmentation across nodes. Allow requests to wait for an in-flight
refill for a bounded duration.

Return advisory Retry-After values when capacity is exhausted, and apply
global backoff when lease acquisition fails or returns an invalid
response.

After serving admitted waiters from new lease grants, restore only
unused grant capacity that remains above the local target, together with
capacity consumed locally but ultimately unused. Perform restorations to
PostgreSQL on a best-effort basis, including during shutdown.
```


cmd/apis_server_test.go
zapateo, line 1: There are various documentations added to this file; are they necessary for the purpose of this PR, or are they something extra that we could do separately?

cmd/workspaces.go
```diff
@@ -148,7 +148,10 @@ func (workspace workspace) CreateEventListener(_ http.ResponseWriter, r *http.Re
 
 // Delete deletes the current workspace with all its connections.
 func (workspace workspace) Delete(_ http.ResponseWriter, r *http.Request) (any, error) {
-	ws, err := workspace.workspace(r)
+	if err := validateForbiddenBody(r); err != nil {
```
zapateo: Some calls to `validateForbiddenBody` have been added. Are they related to the modification of this PR?

tools/errors/errors.go
```diff
@@ -138,30 +139,60 @@ func (e *NotFoundError) WriteTo(w http.ResponseWriter) error {
...
+
+// Error implements the error interface.
+func (e *TooManyRequestsError) Error() string {
+	return e.Message
+}
+
+// WriteTo implements the ResponseWriterTo interface.
+func (e *TooManyRequestsError) WriteTo(w http.ResponseWriter) error {
+	if e.RetryAfter > 0 {
+		seconds := (e.RetryAfter + time.Second - 1) / time.Second
```
zapateo: I'm having trouble understanding this calculation of seconds. The [Seconds](https://pkg.go.dev/time#Duration.Seconds) method already exists, so why this calculation? I need to understand it, and then maybe an explanatory comment would be helpful.

cmd/organization.go
zapateo, line 1: I see that in this file we have the `organization.SetStatus` method, which doesn't appear to be modified by this PR; I don't know if there are other methods for which the same applies.

If so, why isn't the API usage being incremented? If so, is it a deliberate choice or a lack of this PR?

cmd/api.go
```diff
@@ -102,28 +102,53 @@ func (api api) ChangeMemberPasswordByToken(_ http.ResponseWriter, r *http.Reques
...
+// organization.
+type organizationLimits struct {
+	Members     *int `json:"members"`
+	AccessKeys  *int `json:"accessKeys"`
+	Workspaces  *int `json:"workspaces"`
+	Connectors  *int `json:"connectors"`
+	Connections *int `json:"connections"`
+	Pipelines   *int `json:"pipelines"`
+	Rates       *struct {
+		Organization *struct {
```
zapateo: Looking at the limits described here, including their names, it seemed obvious (at least, to me) that they are structured hierarchically, that is, the organization limits include the workspace limits, and the workspace limits include the event limits:

Image

However, looking at the implementation and the code comments:

https://github.com/krenalis/krenalis/blob/93ae8520306ae98b87e8a63e4c2e48577ea18250/core/organization.go#L92-L96

The limit mechanism seems to be quite the opposite, as the limits are disjoint from each other:

Image

I fear could create problems since one believes one can set limits at the organizational level to limit workspaces (and therefore think in a hierarchical manner) when one isn't limiting operations on the workspaces.

Wasn't there any talk of using a term like "non-specific", or something similar, to denote these limits? Furthermore, even if we had a term for "non-workspace limits", to refer to what is currently called "organization" limits, how would we deal with the fact that the limits on the workspace do not include the limits on events, even though the events are imported into the workspace?

I'd like to discuss about this.

core/internal/state/ratelimiter/doc.go
```diff
@@ -0,0 +1,201 @@
...
+//
+// PostgreSQL stores authoritative capacity and is the only shared coordination
+// mechanism between application nodes. The limiter batches organization,
+// workspace, and event requests. PostgreSQL locks the corresponding rows and
+// subtracts granted capacity before returning it. A process crash can lose an
+// unused lease but cannot create additional capacity.
+//
+// Leased capacity is deducted from PostgreSQL before it is consumed locally.
+// Because PostgreSQL may refill while earlier leases remain unused,
+// BurstCapacity bounds the authoritative bucket and each local target, but does
```
zapateo: The term "burst capacity" is used, but in our model, if I understand correctly, this doesn't seem to limit the burst of API requests.

Having multiple nodes, each with its own local buckets, they can respond to a series of API calls, allowing a burst higher than this value.

In fact, here in the documentation it was necessary to clarify that:

> but does not strictly bound the total number of operations performed within a short time interval.

almost as if to suggest that the name could be misleading.

In fact, if I'm not mistaken, this value describes more the "bucket capacity," i.e., the maximum size of the bucket, and a term that goes along these lines seems more intuitive and suited to our model.

Does this make sense to you, or am I missing something? What do you think?

cmd/api.go
zapateo, line 1: Some APIs that impose limits return headers that help the client regulate calls.

For example:

* [GitHub](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10#checking-the-status-of-your-rate-limit)
* [HubSpot](https://developers.hubspot.com/docs/developer-tooling/platform/usage-guidelines#rate-limit-error-responses)

So I wondered:

* Could we have something like this in Krenalis?
* Could we get some similar values, based on how the limits mechanism is currently set up?
* Could it be useful?
* If so, would it be worth it?

We can discuss about this.

core/internal/initdb/rate_limiter_leases.sql
zapateo, line 1: As discussed verbally, the logic of this stored procedure needs to be analyzed more carefully to determine if there are any issues. So, I'm including this adding as a reminder; we'll leave it open and return to it during a second review cycle.

For now, therefore, there is no need for you to make any further changes regarding this, nor for you to respond to this conversation.

core/internal/state/ratelimiter/limiter.go
zapateo, line 1: As discussed verbally, the logic of this package needs to be analyzed more carefully to determine if there are any issues. So, I'm including this adding as a reminder; we'll leave it open and return to it during a second review cycle.

For now, therefore, there is no need for you to make any further changes regarding this, nor for you to respond to this conversation.

zapateo (changes requested):

cmd/workspaces.go
gazerro: Fixed in commit [Remove duplicate request body validation](https://github.com/krenalis/krenalis/pull/2382/changes/7e843de29b78b6bd2e516d064870e73d1f6dad51).

tools/errors/errors.go
gazerro: It is a simple round-up. `Duration.Seconds` returns the number of seconds as a `float64`; it would still need to be rounded up to the next integer, so it would not be sufficient on its own.

cmd/organization.go
gazerro: `SetStatus` is not covered by this rate limiter because it is not authenticated using an organization API key or an event key. Instead, it uses the API key intended for organization-management endpoints.

This rate limiter applies only to endpoints that an organization can access directly. Like the other organization-management methods, `SetStatus` cannot be invoked by an organization itself.

That said, we could also rate-limit those management endpoints by using a separate shared bucket specifically for them.

cmd/api.go
gazerro: Review the API documentation in pull request https://github.com/krenalis/krenalis.com/pull/51, as it should already be clearer there. If it is, we can simply use the same wording in the code comments.

core/internal/state/ratelimiter/doc.go
gazerro: In the public API, I would use `maxCapacity` because it describes what users need to configure without requiring them to understand the internal model. In the implementation, I would use `bucketCapacity` because it precisely identifies the bucket's maximum capacity. What do you think?

cmd/api.go
gazerro: We already return the standard `Retry-After` header with `429 Too Many Requests` responses whenever we can calculate a meaningful retry duration. It is set here: https://github.com/krenalis/krenalis/blob/7e843de29b78b6bd2e516d064870e73d1f6dad51/tools/errors/errors.go#L169

We could also return headers that describe the configured limits for the resource used by the request, such as `ratePerMinute` and `burstCapacity`. We could also add a header similar to GitHub's `X-RateLimit-Resource`. In our case, the value would be `organization`, `workspace`, or `events`.

However, we cannot accurately report the total capacity used or remaining across the whole nodes. Each application node only knows the capacity assigned to its own local lease. It does not know the authoritative total or the capacity assigned to other nodes.

To calculate an exact value, we would need to query PostgreSQL for every request. The leasing mechanism is specifically designed to avoid this. Reporting only the remaining capacity of the local node would therefore be misleading.

A fixed reset timestamp would also not fit our model well. Capacity is refilled continuously instead of being reset at the end of a fixed time window, as is typical of a token-bucket algorithm.

The static limit and resource headers could still be useful for client configuration, diagnostics, and observability. However, `Retry-After` remains the most useful information for automatic client throttling.

I think the additional headers could be valuable, but they should probably be implemented as a separate improvement rather than treated as a requirement of the current rate-limiter implementation.

cmd/apis_server_test.go
zapateo: @gazerro

tools/errors/errors.go
zapateo: As discussed verbally, please consider using [`math.Ceil`](https://pkg.go.dev/math#Ceil), which is also clearer to read.

cmd/organization.go
zapateo: > That said, we could also rate-limit those management endpoints by using a separate shared bucket specifically for them.

As discussed verbally, this might be a good idea.

core/internal/state/ratelimiter/doc.go
zapateo: As discussed verbally, this might be a good idea.

cmd/api.go
zapateo: As discussed verbally, we'll see in the future. For now, it's clearly not worth it.

cmd/api.go
zapateo: As discussed verbally, we could use the suffix "-specific" for the names of the various limits (e.g., "organizationSpecific").

This increases the clarity and scope of the limits.

zapateo (changes requested):

cmd/apis_server_test.go
gazerro: Fixed in commit [cmd: remove unrelated API server test comments](https://github.com/krenalis/krenalis/pull/2382/changes/137c9afe80c8ac53d4c8650227600134e42836ce).

See also PR https://github.com/krenalis/krenalis/pull/2392.

tools/errors/errors.go
gazerro: Improved in commit [Improve Retry-After header calculation readability](https://github.com/krenalis/krenalis/pull/2382/changes/17424bd9489c019e8f1df5043ad4d36cd26db6bd).

cmd/organization.go
gazerro: Done in commit [Add rate limiting for the platform management API](https://github.com/krenalis/krenalis/pull/2382/changes/97dc40e8a5733ac498afa2946ce1e255907f9fb1).

core/internal/state/ratelimiter/doc.go
gazerro: Done in commit [Use "max capacity" instead of "burst capacity"](https://github.com/krenalis/krenalis/pull/2382/changes/98821e844abc6e04323082bed906538a60b57d35).

cmd/api.go
gazerro: Done in commit [Append the "Specific" suffix](https://github.com/krenalis/krenalis/pull/2382/changes/220e838e32930079620782b8684615ea4944e9c0).

cmd/endpoints.go
```diff
@@ -29,7 +29,7 @@ func endpoints(s *apisServer) map[string]endpointHandler {
 		"DELETE /events/listeners/{id}":                       workspace.DeleteEventListener,
 		"DELETE /keys/{key}":                                  organization.DeleteAccessKey, /* Admin console only */
 		"DELETE /members/{id}":                                organization.DeleteMember,    /* Admin console only */
-		"DELETE /organizations/{id}":                          organization.Delete,          /* Needs organizations API key */
+		"DELETE /organizations/{id}":                          organization.Delete,          /* Need platform management API key */
```
zapateo: The `s` in `Needs` should be kept, shouldn't it?

core/internal/initdb/schema.sql
```diff
@@ -408,19 +418,25 @@ CREATE TABLE metadata (
...
+    requests_rate_per_minute,
+    requests_burst_capacity
 ) VALUES (
     gen_random_uuid(),
     '\x'::bytea,
     '\x'::bytea,
     '\x'::bytea,
-    '\x'::bytea
+    '\x'::bytea,
+    100,
```
zapateo, line 456: I need to understand a point. We allow 100 requests per minute for the platform management APIs; however, 100 is also the lease size for a single node; so if a platform management request arrives at one node, the other nodes, which may be receiving other forwarded requests, are left empty and return a 429 error, even though only one request has been served.

For organizations and workspaces, the problem isn't apparent because the limits are much higher, and it would only become apparent as the number of nodes increases.

Does this make sense? Or am I missing something?

zapateo (changes requested):

cmd/endpoints.go
gazerro: Fixed in commit [Fix comments](https://github.com/krenalis/krenalis/pull/2382/changes/c906d636a55036f005b265eb15eef855d3c300d6).

core/internal/initdb/schema.sql
gazerro: When a request arrives, Krenalis first uses any capacity available in the local bucket. If more capacity is needed, PostgreSQL calculates how much has been refilled since the last update.

At 100 units per minute, one unit becomes available every 0.6 seconds. If the bucket is empty and a request needs one unit, it may succeed after 0.6 seconds. It may also succeed sooner if the node still has unused local capacity.

That is a real concern with fixed lease sizes: one node could reserve most of the available capacity, leaving little or none for the others.

The commit [Adapt rate-limit lease sizes to reduce capacity fragmentation](https://github.com/krenalis/krenalis/pull/2382/changes/c6b1c7e74b0f0fba8c6aa194a8cb3ea59c8ad46f) reduces this problem by giving each node a small lease at first. The lease grows only when that node keeps receiving traffic.

So, for the 100-unit platform limit, one request should no longer cause a node to reserve the entire bucket. Other nodes are more likely to find capacity available. Fragmentation can still happen, but it should be much less common.

core/internal/initdb/rate_limiter_leases.sql
zapateo: After various changes made and discussed, this point now seems clear. I'm marking the conversation as resolved.

core/internal/state/ratelimiter/limiter.go
zapateo: After various changes made and discussed, this point now seems clear. I'm marking the conversation as resolved.

core/internal/state/ratelimiter/limiter.go
```diff
@@ -0,0 +1,605 @@
...
+// condition.
+var ErrLimiterUnavailable = errors.New("rate limiter unavailable")
+
+// CapacityExceededError is returned when the requested capacity is unavailable.
+type CapacityExceededError struct {
+	RetryAfter time.Duration
+}
+
+func (err CapacityExceededError) Error() string {
+	return "requested rate-limit capacity is unavailable"
```
zapateo: This error is returned externally by the APIs, for example:

```
GET v1/organizations/F2oxRUvMQCgk: unexpected status code 429: {"error":{"code":"TooManyRequests","message":"requested rate-limit capacity is unavailable"}}
```

But it's at too low a level of abstraction and technicality. It should be reformulated.

test/krenalistester/call.go
zapateo, line 1: The `tryCall` function should be modified to take HTTP 429 errors and Retry-After headers into account.

core/internal/state/state.go
```diff
@@ -1738,3 +1843,21 @@ const (
 	UpdateOnly     ExportMode = "UpdateOnly"
 	CreateOrUpdate ExportMode = "CreateOrUpdate"
 )
+
+// registerRateLimiterCounter registers a counter in the default Prometheus
+// registry. If a counter with the same name is already registered, it returns
+// the existing counter. It returns nil when registration fails or the existing
+// collector is not a counter.
+func registerRateLimiterCounter(options prometheus.CounterOpts) prometheus.Counter {
```
zapateo: This function directly calls `github.com/prometheus/client_golang/prometheus`.

However, we already have the `tools/prometheus` package in Krenalis, which has a function:

https://github.com/krenalis/krenalis/blob/deb2253c67e685472f43a8a72a44cd38fa311202/tools/prometheus/counter.go#L38

Could we actually use the one we already have, or is there really a need to implement a separate function?

And if it were indeed separate, why not put it in `tools/prometheus`?

cmd/authentication_test.go
```diff
@@ -0,0 +1,103 @@
...
+		err := authenticated.applyRateLimitTo(context.Background(), organization, 3)
+		if err != nil {
+			t.Fatalf("apply rate limit: %v", err)
+		}
+		if organization.calls != 1 || organization.cost != 3 {
+			t.Fatalf("expected organization consumption with cost 3, got calls=%d cost=%d", organization.calls, organization.cost)
+		}
+	})
+
+	t.Run("consumes the selected workspace budget", func(t *testing.T) {
```
zapateo: This test appears identical to the previous one, a few lines above. The only difference is the variable name (`workspace` instead of `organization`).

Is this intentional?

cmd/endpoints.go
zapateo, line 1: Currently, when sending a batch of events to Krenalis whose size exceeds the organization's bucket capacity for events, the APIs respond with an HTTP 429 error, when in reality, such a request will never be accepted, either now or in the future, unless the Krenalis configuration is changed.

As discussed verbally, one option is to require a minimum event bucket capacity that allows any batch request to be imported, thus completely eliminating the condition described in the above paragraph.

Please note that the following test has been added to this PR:

https://github.com/krenalis/krenalis/blob/220e838e32930079620782b8684615ea4944e9c0/test/event_rate_limiter_test.go#L16-L18

This reinforces the behavior described above; therefore, this test will need to be revised.

core/internal/collector/decoder.go
```diff
@@ -207,6 +215,8 @@ func (d *decoder) Reset(r *http.Request) error {
 	d.writeKey = ""
 	d.connectionId = ""
 	d.context = nil
+	d.typ = ""
```
zapateo: Is this line, `d.typ = ""`, a change that is necessary to this branch, or is it a separate fix that should be done separately?

test/krenalistester/support.go
```diff
@@ -699,13 +722,13 @@ func (k *Krenalis) SendEvent(writeKey string, message analytics.Message) {
 	if err != nil {
 		k.t.Fatalf("cannot enqueue event: %s", err)
 	}
-	err = client.Close()
+	err = <-cb.ch
```
zapateo, line 731: Isn't this fix, which we were discussing verbally, something that's independent of this change and should be fixed on `main`?

Even if that were the case, I'm fine with leaving it here, to avoid opening another PR. As you prefer.

core/internal/state/ratelimiter/limiter.go
gazerro: Fixed in commit [Clarify rate-limit error message](https://github.com/krenalis/krenalis/pull/2382/changes/4f87f0fa588ad6b9b1d41f96658ccaea0ab1ae51).

test/krenalistester/call.go
gazerro: Done in commit [Handle Retry-After in tryCall](https://github.com/krenalis/krenalis/pull/2382/changes/3c0573ef404db09c576458d445201e06496c3c6f).

Note that I capped the Retry-After value returned by the rate limiter at 20 minutes, so clients and tests never have to wait an excessively long time before retrying.

core/internal/state/state.go
gazerro: Done in commit [Use shared Prometheus counters for rate limiter metrics](https://github.com/krenalis/krenalis/pull/2382/changes/4049bba601a6bfb7ed37554a5e6b1e72d6444590).

cmd/authentication_test.go
gazerro: Fixed in commit [Remove duplicate rate-limit test](https://github.com/krenalis/krenalis/pull/2382/changes/5f58a506d7df91dab66efd9deb7cc6125bc1d365).

cmd/endpoints.go
gazerro: Done in commit [Enforce minimum event capacity for ingestion batches](https://github.com/krenalis/krenalis/pull/2382/changes/0c16d1a8444b15df24f0dcdce0d9f0a40181cb26).

core/internal/collector/decoder.go
gazerro: Done in commit [Remove d.typ = ""](https://github.com/krenalis/krenalis/pull/2382/changes/ae61531f5a8adf528953c824c31692ce3632d5b8).

See PR #2402.

test/krenalistester/support.go
gazerro: The issue does not occur on `main` because the request is never retried. In this branch, however, it can be retried and would fail without this change.

zapateo (changes requested):
Please, fix the conflicts with `main`.

core/internal/state/state.go
```diff
@@ -181,9 +214,12 @@ func (state *State) Account(id int) (*Account, bool) {
 
 // Close closes the state. When it is called, no calls to the State methods
 // should be in progress, and no further calls should be made.
-func (state *State) Close() {
+func (state *State) Close(ctx context.Context) {
 	state.close.cancel()
 	state.close.Wait()
+	state.rateLimiter.Close(ctx)
+	// Limiter.Close reads unused capacity from buckets still owned by the state.
+	runtime.KeepAlive(state)
```
zapateo, line 223: Please, review this code and check if there is an alternative.

zapateo (changes requested):

core/internal/state/state.go
gazerro: See commit [Clarify why runtime.KeepAlive(state) is needed](https://github.com/krenalis/krenalis/pull/2382/changes/ffdf37682eda02e7fd1d1acc8246509e428a0c45).

zapateo (approved):

PR #2384, opened 2026-07-29
core/internal/transformers/lambda: add mutex to protect `client`

zapateo:
## Commit message

```
core/internal/transformers/lambda: add mutex to protect `client` (#2384)

Also improve documentation of Close method to clarify its use.
```

### DIFF
--- a/core/internal/transformers/lambda/lambda.go
+++ b/core/internal/transformers/lambda/lambda.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/krenalis/krenalis/core/internal/state"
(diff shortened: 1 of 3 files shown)


core/internal/transformers/lambda/lambda.go
```diff
@@ -32,6 +33,7 @@ import (
 
 type function struct {
 	settings Settings
+	mu       sync.Mutex // for client.
```
gazerro: Move this comment to the client field, for example:

`client   *lambda.Client // protected by mu`

gazerro (changes requested):

core/internal/transformers/lambda/lambda.go
zapateo: Done in [Move comment from mu to client](https://github.com/krenalis/krenalis/pull/2384/commits/eafe302cf19c66acc2d5b007cbe0683932b497f8).

gazerro (approved):

PR #2386, opened 2026-07-29
core/internal/collector/sender: fix a race in the sender test

gazerro:
```
core/internal/collector/sender: fix a race in the sender test (#2386)

The test read app.iteration without holding the mutex that protects it,
causing a data race. It also checked the counter after incrementing it,
allowing concurrent calls to wrap it to zero before the check and
leaving the overflow undetected.
```

### DIFF
--- a/core/internal/collector/sender/sender_test.go
+++ b/core/internal/collector/sender/sender_test.go
@@ -795,7 +795,7 @@ type application struct {
 	seed uint64
 
 	mu        sync.Mutex
-	iteration uint64
+	iteration uint64   // protected by mu
 	n         int      // protected by mu
 	consumed  []string // ids of the consumed events; protected by mu
 	sends     []send   // protected by mu
--- a/core/internal/collector/sender/sender_test.go
+++ b/core/internal/collector/sender/sender_test.go
@@ -844,13 +844,13 @@ func (app *application) SendEvents(ctx context.Context, events connectors.Events
 	// Get the current iteration number.
 	var iteration uint64
 	app.mu.Lock()
+	if app.iteration == math.MaxUint64 {
+		app.mu.Unlock()
+		panic("iteration is out of range")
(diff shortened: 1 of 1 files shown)


core/internal/collector/sender/sender_test.go
```diff
@@ -844,13 +844,13 @@ func (app *application) SendEvents(ctx context.Context, events connectors.Events
 	// Get the current iteration number.
 	var iteration uint64
```
zapateo, line 845: While not strictly related to this change, we could avoid declaring this variable here (as it adds a bit of confusion when reading the test code) and assign it directly using `:=` when needed.

Am I wrong?

It's not important, I just wanted to point it out. You decide if it's worth it.

core/internal/collector/sender/sender_test.go
zapateo, line 1: The `iteration` field has no documentation about being protected by `mu`:

https://github.com/krenalis/krenalis/blob/3457a7c11f131581daad2e31d9fb41dc7e9ce31a/core/internal/collector/sender/sender_test.go#L793-L802

Is there a reason? I'd be in favor of adding the comment.

It could also be (I don't know) that the lack of such comment may have contributed in part to causing the race.

zapateo (changes requested):

core/internal/collector/sender/sender_test.go
gazerro: As discussed, I'll leave it as it is.

core/internal/collector/sender/sender_test.go
gazerro: Done in commit [Add missing comment](https://github.com/krenalis/krenalis/pull/2386/changes/ff5437e7261c9424e1628e19572e64d3d7853209).

zapateo (approved):

PR #2395, opened 2026-08-02
core/internal/streams/nats: prevent premature event redelivery

### LINKED ISSUE #2369: Pipeline metrics incorrectly repeat Transform and Validate counts across minute buckets
For a Dummy destination connection whose event dispatches fail, the per-minute pipeline metrics continuously report the same non-zero counts for the **Transform** and **Validate** steps in every subsequent minute bucket.

These counts persist indefinitely, including in buckets created long after the events were originally ingested. Metrics for the other pipeline steps behave correctly: their counts appear only in the relevant minute bucket and eventually leave the 15-minute window.

As a result, the pipeline funnel on the connection’s Metrics page displays inconsistent totals. For example, it may show 70 events received but more than 1,200 events transformed. The values displayed on the **Transform → Validate** and **Validate → Send** arrows keep increasing and then oscillate between values such as 1024, 1122, 1081, 1090, and 1213, instead of eventually returning to zero.

The issue can also be observed in the browser console by inspecting responses from the following endpoint:

`v1/pipelines/metrics/minutes/{minutes}`

A few minutes after ingestion, the `passed` array may look like this:

```text
idx 0..10: [0, 0, 0, 0, 0, 0]
idx 11:    [38, 0, 38, 62, 62, 0]  <- minute in which the events were sent
idx 12:    [ 0, 0,  0, 76, 76, 0]
idx 13:    [ 0, 0,  0, 76, 76, 0]
idx 14:    [ 0, 0,  0, 76, 76, 0]
```

Approximately 20 minutes later, the original ingestion minute should have completely left the 15-minute window. However, the same Transform and Validate counts are still repeated across the buckets:

```text
idx 0..13: [0, 0, 0, 76, 76, 0]
idx 14:    [0, 0, 0, 12, 12, 0]
```

Image

### How to reproduce

1. Create a Dummy destination with at least one enabled event pipeline, such as **Send generic event**.
2. Configure Dummy so that event dispatch fails. For example, configure the **URL to which event dispatch POSTs are made** setting to a local URL that does not respond.
3. Link a source to Dummy, such as JavaScript, and send a batch of events.
4. Open the **Metrics** page of Dummy and monitor the **Pipeline** section.

### Expected behavior

Each pipeline step count should be attributed only to the minute in which that processing occurred.

Once the relevant minute leaves the 15-minute window, its counts should no longer be included in the endpoint response or in the Pipeline funnel totals.

gazerro:
```
core/internal/streams/nats: prevent premature event redelivery (#2395)

Events are acknowledged only after they have passed through the
processing pipeline and been dispatched. If processing takes longer than
the NATS acknowledgment timeout, NATS assumes that the delivery failed
and delivers the event again, causing some processing steps to run more
than once.

Keep the acknowledgment deadline from expiring while an event is being
processed by sending periodic progress notifications. Stop these
notifications when the event is acknowledged or the consumer is closed,
and ensure that the final acknowledgment is sent only once.

Integrate the per-destination acknowledgments introduced by #2406 under
the same tracked NATS message. Continue progress notifications after an
individual destination completes, and stop only after the final
destination completes or the consumer is closed.

Make the acknowledgment timeout configurable through
KRENALIS_NATS_ACK_WAIT and AWS Parameter Store, with a safe default and
minimum value. Keep MaxDeliver explicitly unlimited until a terminal
retry policy is defined.

Fixes #2369
```


cmd/config.go
```diff
@@ -459,6 +459,17 @@ func loadConfig(ctx context.Context, source string) (*Config, error) {
 		}
 		settings.NATS.NKey = ed25519.NewKeyFromSeed(seed)
 	}
+	settings.NATS.AckWait = natsopts.DefaultAckWait
+	if wait := conf.Get("KRENALIS_NATS_ACK_WAIT"); wait != "" {
```
zapateo, line 463: The new variable `KRENALIS_NATS_ACK_WAIT` should be documented on krenalis.com, shouldn't it?

Also see https://github.com/krenalis/krenalis/pull/2395/changes#r3711144453.

cmd/internal/config/aws/aws.go
```diff
@@ -363,6 +363,7 @@ var parameters = map[string]string{
 	"/max-queued-events-per-destination":      "MAX_QUEUED_EVENTS_PER_DESTINATION",
 	"/maxmind-db-path":                        "MAXMIND_DB_PATH",
 	"/member-email-from":                      "MEMBER_EMAIL_FROM",
+	"/nats/ack-wait":                          "NATS_ACK_WAIT",
```
zapateo, line 366: `/nats/ack-wait` should be documented on krenalis.com, shouldn't it?

Also see https://github.com/krenalis/krenalis/pull/2395/changes#r3711133212.

core/internal/streams/nats/nats.go
zapateo, line 1: Please also check the aspects discussed verbally.

zapateo (changes requested):

core/internal/streams/nats/nats.go
gazerro: I've merged `main` into the current branch, including the [core: track acknowledgments per event destination](https://github.com/krenalis/krenalis/commit/1ae8e35bc5aa575ca57bd9df03e3fc0aec36c6ba) commit.

cmd/config.go
gazerro: See https://github.com/krenalis/krenalis.com/pull/52.

cmd/internal/config/aws/aws.go
gazerro: See https://github.com/krenalis/krenalis.com/pull/52.

core/internal/streams/nats/nats.go
gazerro: See commit [Clear heartbeat snapshots on cancellation](https://github.com/krenalis/krenalis/pull/2395/changes/5786f24b80b7054c72dc8abf11682425dcfcceee).

gazerro: See commit [Reuse pipeline processing step constants in tests](https://github.com/krenalis/krenalis/pull/2395/changes/3aec8d8e4e1295b3a46d1a551224434be1575747).

test/dispatch_events_to_dummy_test.go
```diff
@@ -17,7 +17,15 @@ import (
 	"github.com/krenalis/analytics-go"
 )
 
+const (
+	pipelineProcessingSteps = 6
```
zapateo: Shouldn't this constant value be `7` now?

test/dispatch_events_to_dummy_test.go
```diff
@@ -75,12 +83,59 @@ func TestDispatchEventsToDummy(t *testing.T) {
...
 
+// waitPipelinePassedMetrics waits until processing metrics for pipelineID are
+// visible and returns their totals across all minute buckets.
+func waitPipelinePassedMetrics(t *testing.T, k *krenalistester.Krenalis, pipelineID string) [pipelineProcessingSteps]int {
+	t.Helper()
+	poll := time.NewTicker(100 * time.Millisecond)
+	defer poll.Stop()
+	timeout := time.NewTimer(5 * time.Second)
+	defer timeout.Stop()
+	var totals [6]int
```
zapateo: See [this other note](https://github.com/krenalis/krenalis/pull/2395/changes#r3804054353).

cmd/config.go
```diff
@@ -459,6 +459,17 @@ func loadConfig(ctx context.Context, source string) (*Config, error) {
...
+		ackWait, err := time.ParseDuration(wait)
+		if err != nil {
+			return nil, fmt.Errorf("invalid duration value specified for KRENALIS_NATS_ACK_WAIT: %s", err)
+		}
+		if ackWait < natsopts.MinAckWait {
+			return nil, fmt.Errorf("KRENALIS_NATS_ACK_WAIT must be at least %s", natsopts.MinAckWait)
+		}
+		settings.NATS.AckWait = ackWait
```
zapateo: `settings.NATS.AckWait` has a lower bound but no upper bound. Is this intentional?

core/internal/streams/nats/nats.go
```diff
@@ -665,16 +648,201 @@ func (batch *batch) Publish(ctx context.Context, topics []string, event map[stri
...
+	if a.remaining != 0 {
+		a.mu.Unlock()
+		return
+	}
+	a.done = true
+	a.mu.Unlock()
+
+	a.manager.Remove(a)
+	if err := a.msg.Ack(); err != nil {
+		slog.Warn(fmt.Sprintf("nats: cannot ack event: %s", err))
```
zapateo: Here, logging is unstructured (it is just a string), while in other places, such as this one:

https://github.com/krenalis/krenalis/blob/5786f24b80b7054c72dc8abf11682425dcfcceee/core/internal/streams/nats/nats.go#L738

Logging is structured. Is it worth standardizing?

zapateo (changes requested):

test/dispatch_events_to_dummy_test.go
gazerro: Fixed in commit [Fix test after merge](https://github.com/krenalis/krenalis/pull/2395/changes/f493a7e30628fd52a81fe87b65d9bf1f34a937f7).

test/dispatch_events_to_dummy_test.go
gazerro: Fixed in commit [Fix test after merge](https://github.com/krenalis/krenalis/pull/2395/changes/f493a7e30628fd52a81fe87b65d9bf1f34a937f7).

cmd/config.go
gazerro: AckWait è documentato in NATS come:

```
// AckWait defines how long the server will wait for an acknowledgement
// before resending a message. If not set, server default is 30 seconds.
AckWait time.Duration `json:"ack_wait,omitempty"`
```

As we discussed, having a minimum is more important than having a maximum.

core/internal/streams/nats/nats.go
gazerro: Fixed in commit [Fix slog.Warn call](https://github.com/krenalis/krenalis/pull/2395/changes/7a002000f6537bb4540f5949a1e63772aec47762).

test/dispatch_events_to_dummy_test.go
```diff
@@ -17,7 +17,15 @@ import (
...
 )
 
+const (
+	pipelineProcessingSteps = 7
+	transformationStep      = 4
+	outputValidationStep    = 5
+)
+
 func TestDispatchEventsToDummy(t *testing.T) {
+	const ackWait = time.Second
```
zapateo: Setting an environment variable this way only works if, during testing, Krenalis is started as an external process (since it is then provided with the test process environment):

https://github.com/krenalis/krenalis/blob/eba694ddb96cf4e904b26659997730248f3141cf/test/krenalistester/krenalistester.go#L337

But if Krenalis is started "embedded" within the test process (the test “debugging mode“), the environment variables aren't read at all, as the configuration is passed by directly initializing a structure. `cmd.Config`:

https://github.com/krenalis/krenalis/blob/eba694ddb96cf4e904b26659997730248f3141cf/test/krenalistester/krenalistester.go#L377

This means that the behavior of tests in one case is different from that of tests in another.

One option is to globally set a constant value for _KRENALIS_NATS_ACK_WAIT_ setting for all tests, setting it directly in `test/krenalistester/krenalistester.go` (in `Krenalis.Start`), in both cases shown above (both when Krenalis is started as an external process and when it is embedded).

If, however, it is necessary to configure it at the test level, the test infrastructure must be modified to expose a way to set this value (for example, as we already do something similar with [Krenalis.PopulateProfileSchema](https://github.com/krenalis/krenalis/blob/eba694ddb96cf4e904b26659997730248f3141cf/test/krenalistester/krenalistester.go#L164-L166)).

zapateo (changes requested):

test/dispatch_events_to_dummy_test.go
gazerro: Done in this commit [Configure NATS acknowledgment wait](https://github.com/krenalis/krenalis/pull/2395/changes/baf6a3da472246a1d3bc78a93b60fc23824f6b1b).

This commit also queries two minutes of pipeline metrics because transformation and delivery can fall on opposite sides of a UTC minute boundary.

Note that `TestDispatchEventsToDummy` fails when run in-process until the fix introduced by PR #2425 is applied.

zapateo (approved):

PR #2396, opened 2026-08-02
core/internal/streams/nats: prevent prefetched event redelivery

gazerro:
```
core/internal/streams/nats: prevent prefetched event redelivery (#2396)

JetStream starts the acknowledgment timer when a message is fetched. The
client may fetch a batch while Krenalis is still processing an earlier event,
so messages waiting behind it can use much of their acknowledgment time before
processing begins. If they wait long enough, JetStream redelivers them.

Fetch messages in bounded batches and start acknowledgment tracking before
each message enters the processing queue. Fetch another batch only when the
queue has enough room for every message that may be returned. This applies
backpressure without reducing the normal batch size and preserves per-shard
processing order.

Stop tracking and negatively acknowledge messages that remain buffered when
the consumer closes. Add regression tests proving that fetched messages receive
progress heartbeats while decoding is blocked and that fetching resumes as
processing frees capacity.
```

### DIFF
--- a/core/internal/streams/nats/nats.go
+++ b/core/internal/streams/nats/nats.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"iter"
 	"log/slog"
 	"slices"
 	"strconv"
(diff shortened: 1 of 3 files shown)


zapateo (changes requested):
Evaluate the things discussed verbally about this branch.

zapateo (changes requested):
Please, see the points discussed verbally.

gazerro:
> Please, see the points discussed verbally.

See the document I gave you.

core/internal/streams/nats/nats.go
```diff
@@ -447,138 +448,56 @@ func (s *stream) Batch(ctx context.Context) (streams.BatchPublisher, error) {
...
 // consumer implements the streams.Consumer interface.
 type consumer struct {
 	stream *stream
+	topic  string
 	events chan streams.Event
-	cancel context.CancelFunc
 	acks   *ackManager
+	close  struct {
+		cancel context.CancelFunc // stops shard consumption
+		closed atomic.Bool        // indicates whether Close has been called
```
zapateo: As discussed verbally, I would consider using a simple boolean instead of an `atomic.Bool`, since the current implementation would cause unexpected behavior for the `Close` in the event of a concurrent call (the second `Close` would return immediately when the first call to the _“real”_ `Close` has not yet, in fact, closed).

With the boolean, we still guarantee idempotence, but we panic in the event of an (accidental, unplanned) concurrent call. This seems to be the best solution.

zapateo (changes requested):

core/internal/streams/nats/nats.go
gazerro: Done in commit [Simplify consumer close state](https://github.com/krenalis/krenalis/pull/2396/changes/d21edddfa51b3eafb3eb0e68e1b3e68151705ee7).

zapateo (approved):

PR #2398, opened 2026-08-02
core/internal/streams/nats: rate-limit acknowledgment error logs

gazerro:
```
core/internal/streams/nats: rate-limit acknowledgment error logs  (#2398)

A NATS outage can cause many final message acknowledgments to fail in a
short period, producing a large number of identical log entries.

Log the first acknowledgment error immediately, then rate-limit repeated
instances of the same error while keeping track of their total number.
Log different errors immediately and report when acknowledgment delivery
succeeds again after repeated failures.
```


zapateo (changes requested):
Please check the document I sent you separately.

gazerro:
> Please check the document I sent you separately.

Done in commit [Keep acknowledgment errors rate-limited](https://github.com/krenalis/krenalis/pull/2398/changes/4c9751b9d735a63a03d97d82a02355bb99387f8a).

zapateo (changes requested):
Please check the second document that I shared with you.

gazerro:
Closing without merge. This layer is being removed from the stack; acknowledgment log rate limiting may be revisited later as a separately designed change.

PR #2399, opened 2026-08-02
connectors, core: add ordering groups to event types

gazerro:
```
connectors, core: add ordering groups to event types (#2399)

The sender currently preserves the order in which events for the same user are
delivered across all event pipelines of a destination connection. This
guarantees ordering, but also couples pipelines that may not depend on each
other.

If delivery for one pipeline cannot complete, later events for the same user
cannot proceed through the other pipelines of that connection, even when those
pipelines could deliver them successfully.

Allow connectors to define which event types must share this ordering
constraint. Event types in the same ordering group will eventually share the
same per-user delivery stream, while event types in different groups will not
share ordering constraints.

This commit:

 * Adds OrderingGroup to connector event types.
 * Uses the event type ID as the ordering group when none is explicitly set.
 * Places event type IDs and ordering groups in the same namespace.
 * Applies the property-name syntax and a maximum length of 25 characters to
    both identifiers.
 * Assigns ordering groups to existing connectors.
 * Resolves and stores the ordering group when a pipeline is created, avoiding
    the need to load connector event type definitions during initialization.
 * Backfills ordering groups for existing pipelines and adds the corresponding
   database constraints.

Future changes will use the stored ordering groups to separate event delivery
streams. This commit does not change event routing or delivery behavior.
```


connectors/skills/create-application/references/events.md
zapateo: The documentation on krenalis.com should be updated as well. For example, the page https://www.krenalis.com/docs/create-integration/application/send-events.

cmd/internal/mcp/tools.go
zapateo: This change doesn't seem to be linked to this PR... I don't know—maybe we need to merge 'main' into this branch? I'm not sure.

connectors/applications.go
```diff
@@ -689,3 +701,12 @@ type ReceivedEventContextSession interface {
 	ID() (int, bool)
 	Start() (bool, bool)
 }
+
+// OrderingGroup returns the effective ordering group of an event type. If no
+// group is set, it returns the event type ID.
```
zapateo: Based on the way this documentation is phrased, it appears that the function could return two distinct things:

* the Ordering Group
* the Event Type ID

However, if I understand correctly (and judging by the commit message and the function name, 'OrderingGroup'), this function always returns an Ordering Group; in certain cases, if the ordering group value has not been explicitly set by the connector, the event type ID is used as the value, but conceptually it is still an ordering group.

Is that correct? If so, do you think the documentation could be rephrased to clarify this point?

core/internal/connections/httpclient/rate_limiter_test.go
zapateo: As noted elsewhere, this change does not seem related to this PR; perhaps a merge from `main` into this branch is needed.

core/internal/initdb/upgrade.go
zapateo: It seems that the position of the 'pipelines.ordering_group' column differs depending on whether the database is created from scratch or updated.

However, to conduct a "real" comparison test against the main branch, it would be best to merge main into this branch; then we can actually verify whether the updater performs as expected.

So, perhaps you could do two things:

1. Merge main into this branch
2. Check if the issue with the 'ordering_group' position still persists

warehouses/postgresql/internal/readonlysql/validate_test.go
zapateo: As noted elsewhere, this change does not seem related to this PR; perhaps a merge from 'main' into this branch is needed.

warehouses/snowflake/checkreadonlyaccess_test.go
zapateo: As noted elsewhere, this change does not seem related to this PR; perhaps a merge from 'main' into this branch is needed.

core/internal/initdb/schema.sql
```diff
@@ -188,7 +188,11 @@ CREATE TABLE pipelines (
     id varchar(12) NOT NULL CHECK (id ~ '^[1-9A-HJ-NP-Za-km-z]{12}$'),
     connection varchar(12) NOT NULL REFERENCES connections ON DELETE CASCADE,
     target pipeline_target NOT NULL,
-    event_type varchar(100) NOT NULL,
+    event_type varchar(25) NOT NULL CONSTRAINT pipelines_event_type_check
```
zapateo: Constraints are defined here, but they are also validated in the Go code.

Doesn't this conflict with what is stated in the AGENTS.md file?

https://github.com/krenalis/krenalis/blob/97da8cd8bd000f40cb11aaf2347187dba50b39cf/AGENTS.md#L190-L193

core/internal/connections/application.go
zapateo: There are differences in how the `EventTypes` (plural) and `EventType` (singular) methods handle certain values ​​returned by the connector:

* a `nil` event type returns an error in `EventTypes` («connector ... returned a nil event type»), whereas it is ignored in `EventType` («candidate == nil... continue...»)
* an invalid ID for other event types is ignored by `EventType`, but returns an error in `EventTypes`
* a duplicate ID returns an error in both functions, but the error message differs (in the case of `EventType`, it does not include the connector name).

Is this behavior intentional, or should it be unified?

core/connection.go
```diff
@@ -395,8 +397,20 @@ func (this *Connection) CreatePipeline(ctx context.Context, target Target, event
 
 	// Only for destination event pipeline checks that the out schema is aligned with the event type's schema.
```
zapateo: This comment now precedes a block that does various things—not just the ones stated—so perhaps it should be expanded? Or maybe it's not worth it since we need to revisit it later anyway, given the TODO? It's up to you.

core/connection.go
```diff
@@ -1454,11 +1465,13 @@ func (this *Connection) PipelineTypes(ctx context.Context) ([]PipelineType, erro
 				}
 				// Destination/Application/Event.
 				for _, et := range eventTypes {
+					orderingGroup := connectors.OrderingGroup(et)
```
zapateo: We could even avoid the local variable 'orderingGroup', couldn't we?

```go
// Destination/Application/Event.
for _, et := range eventTypes {
	pipelineTypes = append(pipelineTypes, PipelineType{
		Name:          et.Name,
		Description:   et.Description,
		Target:        TargetEvent,
		EventType:     new(et.ID),
		OrderingGroup: new(connectors.OrderingGroup(et)),
	})
}
```

zapateo (changes requested):

cmd/internal/mcp/tools.go
gazerro: I merged the latest changes from `main`, and that fixed it.

connectors/applications.go
gazerro: Done in commit [Clarify OrderingGroup comment](https://github.com/krenalis/krenalis/pull/2399/changes/66e2b420210688a72bbefb068317cb6c62807407).

core/internal/connections/httpclient/rate_limiter_test.go
gazerro: I merged the latest changes from `main`, and that fixed it.

warehouses/postgresql/internal/readonlysql/validate_test.go
gazerro: I merged the latest changes from `main`, and that fixed it.

warehouses/snowflake/checkreadonlyaccess_test.go
gazerro: I merged the latest changes from `main`, and that fixed it.

core/internal/initdb/upgrade.go
gazerro: Fixed in commit [Preserve pipeline column order during database upgrades](https://github.com/krenalis/krenalis/pull/2399/changes/7feeaccbbf18e24020d96a9b699a5451346da34b).

core/internal/initdb/schema.sql
gazerro: Done in commit [Remove redundant checks on pipeline event identifiers](https://github.com/krenalis/krenalis/pull/2399/changes/f5159fd0150fec1f933a68b36dbaebde11955bbf).

core/connection.go
gazerro: Done in commit [Avoid the local variable orderingGroup](https://github.com/krenalis/krenalis/pull/2399/changes/e467a0dd0a233ff95d6e59b707bcea2811029b51).

connectors/skills/create-application/references/events.md
gazerro: See https://github.com/krenalis/krenalis.com/pull/64.

core/internal/connections/application.go
gazerro: The first two differences are intentional: `EventTypes` validates the whole list, while `EventType` validates only the requested type. A `nil` entry or an invalid `ID` belonging to another event type therefore does not prevent the lookup from succeeding. I've documented this distinction.

The different duplicate-ID error messages were an inconsistency. I've made them identical, including the connector name.

These differences were introduced in this PR; main does not have the singular `EventType` method. The tests pass.

See commit [Clarify event type validation and unify duplicate errors](https://github.com/krenalis/krenalis/pull/2399/changes/b23425d7c8e774a3d6e8d2ad1ff3f84a40a07adf).

core/connection.go
gazerro: Done in commit [Separate event IDs from ordering groups](https://github.com/krenalis/krenalis/pull/2399/changes/2cbb9a2fabcbf901569773e710c884726a74f4e1).

core/internal/initdb/upgrade.go
zapateo: The ordering now appears correct to me.

However, there is a difference in the `pipelines_required_consents_operator_check` column:

Updated database (main → this branch):

```sql
CONSTRAINT pipelines_required_consents_operator_check CHECK (((required_consents_operator)::text = ANY (ARRAY[('and'::character varying)::text, ('or'::character varying)::text]))),
```

Database created from scratch on this branch:

```sql
CONSTRAINT pipelines_required_consents_operator_check CHECK (((required_consents_operator)::text = ANY ((ARRAY['and'::character varying, 'or'::character varying])::text[]))),
```

I don't know how significant it is, but I'm mentioning it to you anyway.

connectors/applications.go
zapateo: Good. Note that the commit message should be updated.

zapateo (changes requested):

connectors/applications.go
gazerro: Done.

core/internal/initdb/upgrade.go
gazerro: Fixed in commit [Preserve required consents operator constraint during upgrades](https://github.com/krenalis/krenalis/pull/2399/changes/29e88e8f7931b9a9b2b02664f790ed33011cf3f9).

zapateo (approved):

PR #2401, opened 2026-08-03
connectors, core: add delivery endpoints to event types

gazerro:
```
connectors, core: add delivery endpoints to event types (#2401)

Ordering groups define which events for the same user must preserve their
delivery order, but they do not indicate which pipelines share the same
destination endpoint.

A connection may deliver different event types through independent endpoints.
If one endpoint becomes unavailable, future changes can use this information to
suspend only the pipelines that use that endpoint. Conversely, multiple
ordering groups may use the same endpoint and should share its availability
state.

Allow connectors to associate event types with a delivery endpoint. Event types
in the same ordering group must use the same endpoint because suspending either
endpoint independently could violate their shared delivery order.

This commit:

* Adds DeliveryEndpoint to connector event types.
* Uses the empty value as the connector's default endpoint, shared by all event
  types that do not declare an explicit endpoint.
* Applies the property-name syntax and a maximum length of 25 characters to
  explicit endpoint identifiers.
* Verifies that event types in the same ordering group use the same endpoint.
* Resolves and stores the endpoint when a pipeline is created, avoiding the
  need to load connector event type definitions during initialization.
* Makes the endpoint available through pipeline state and the API.
* Backfills existing pipelines with the default endpoint and adds the
  corresponding database constraint.

Future changes will use delivery endpoints to isolate consumption and failure
handling. This commit does not change event routing or delivery behavior.
```

### DIFF
--- a/connectors/applications.go
+++ b/connectors/applications.go
@@ -231,8 +231,8 @@ func (r FailureReason) String() string {
 	panic(fmt.Errorf("unexpected FailureReason %d", r))
 }
 
-// MaxEventTypeIdentifierLen is the maximum length of event type and ordering
-// group identifiers.
+// MaxEventTypeIdentifierLen is the maximum length of event type, ordering
+// group, and delivery endpoint identifiers.
 const MaxEventTypeIdentifierLen = 25
 
 // EventType represents a type of event that can be sent to an application.
@@ -255,6 +255,13 @@ type EventType struct {
 	// the syntax of a property name and cannot be longer than
 	// MaxEventTypeIdentifierLen characters.
 	OrderingGroup string
+
+	// DeliveryEndpoint identifies the destination endpoint used for events of
+	// this type. Event types in the same ordering group must use the same endpoint.
+	// An empty value selects the connector's default endpoint. If set, the value
+	// must follow the syntax of a property name and be no longer than
+	// MaxEventTypeIdentifierLen characters.
+	DeliveryEndpoint string
 
 	// DefaultFilter is the default filter to use for pipelines.
 	DefaultFilter string
--- a/connectors/skills/create-application/references/events.md
+++ b/connectors/skills/create-application/references/events.md
@@ -19,7 +19,12 @@ Use it to choose the right iteration method and payload-building pattern.
 - Set `OrderingGroup` to the same value on event types whose events must remain
   ordered for each user. Ordering groups follow the same syntax and length
   limit as IDs. If `OrderingGroup` is empty, the event type ID is used.
-- Do not change an ID or ordering group after the connector has been released.
+- Set `DeliveryEndpoint` when event types use separate destination endpoints.
+  Event types in the same ordering group must use the same delivery endpoint.
+  An empty value identifies the connector's default endpoint. Explicit endpoint
+  identifiers follow the same syntax and length limit as event type IDs.
+- Do not change an ID, ordering group, or delivery endpoint after the connector
+  has been released.
 - Return only event types the connector actually supports.
 
 ## EventTypeSchema
--- a/core/connection.go
+++ b/core/connection.go
@@ -65,11 +65,12 @@ type Connection struct {
 
 // EventType represents an event type of a destination application connection.
 type EventType struct {
-	ID            string `json:"id"`
-	Name          string `json:"name"`
-	Description   string `json:"description"`
-	OrderingGroup string `json:"orderingGroup"`
-	DefaultFilter string `json:"defaultFilter"`
+	ID               string `json:"id"`
+	Name             string `json:"name"`
+	Description      string `json:"description"`
+	OrderingGroup    string `json:"orderingGroup"`
+	DeliveryEndpoint string `json:"deliveryEndpoint"`
+	DefaultFilter    string `json:"defaultFilter"`
 }
 
 type PipelineInfo struct {
@@ -104,11 +105,12 @@ var dummyGroupsSchema = types.Object([]types.Property{
 
 // PipelineType represents a pipeline type.
 type PipelineType struct {
-	Name          string  `json:"name"`
-	Description   string  `json:"description"`
-	Target        Target  `json:"target"`
-	EventType     *string `json:"eventType"`
-	OrderingGroup *string `json:"orderingGroup"`
+	Name             string  `json:"name"`
+	Description      string  `json:"description"`
+	Target           Target  `json:"target"`
+	EventType        *string `json:"eventType"`
+	OrderingGroup    *string `json:"orderingGroup"`
+	DeliveryEndpoint *string `json:"deliveryEndpoint"`
 }
 
 // AbsolutePath returns the absolute representation of the given path, based
@@ -399,7 +401,7 @@ func (this *Connection) CreatePipeline(ctx context.Context, target Target, event
 
 	// Only for destination event pipeline checks that the out schema is aligned with the event type's schema.
 	// See issue https://github.com/krenalis/krenalis/issues/2086.
-	var orderingGroup string
+	var orderingGroup, deliveryEndpoint string
 	if eventType != "" {
 		app := this.application()
 		et, err := app.EventType(ctx, eventType)
@@ -410,6 +412,7 @@ func (this *Connection) CreatePipeline(ctx context.Context, target Target, event
 			return "", err
 		}
 		orderingGroup = connectors.OrderingGroup(et)
+		deliveryEndpoint = et.DeliveryEndpoint
 		eventTypeSchema, err := app.Schema(ctx, state.TargetEvent, eventType)
 		if err != nil {
 			return "", err
@@ -436,6 +439,7 @@ func (this *Connection) CreatePipeline(ctx context.Context, target Target, event
 		Enabled:            pipeline.Enabled,
 		EventType:          eventType,
 		OrderingGroup:      orderingGroup,
+		DeliveryEndpoint:   deliveryEndpoint,
 		InSchema:           inSchema,
 		OutSchema:          pipeline.OutSchema,
 		RequiredConsents:   toStateRequiredConsents(pipeline.RequiredConsents),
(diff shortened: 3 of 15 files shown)


core/connection.go
```diff
@@ -65,11 +65,12 @@ type Connection struct {
...
-	ID            string `json:"id"`
-	Name          string `json:"name"`
-	Description   string `json:"description"`
-	OrderingGroup string `json:"orderingGroup"`
-	DefaultFilter string `json:"defaultFilter"`
+	ID               string `json:"id"`
+	Name             string `json:"name"`
+	Description      string `json:"description"`
+	OrderingGroup    string `json:"orderingGroup"`
+	DeliveryEndpoint string `json:"deliveryEndpoint"`
```
zapateo: The `DeliveryEndpoint` field should be documented on krenalis.com, in the APIs section.

Consider documenting the `OrderingGroup` field alongside this one as well.

connectors/skills/create-application/references/events.md
zapateo: The documentation on krenalis.com regarding the development of a connector should be updated.

core/internal/connections/application.go
```diff
@@ -112,8 +112,9 @@ func (app *Application) Connector() string {
 
 // EventType returns the application's event type with the specified ID. The
 // returned event type is owned by the connector and must not be modified.
-// It validates only the matching event type and rejects duplicates of its ID;
-// nil entries and other event types are ignored.
+// It validates the matching event type, rejects duplicates of its ID, and
+// verifies that its ordering group uses a single delivery endpoint. Nil entries
```
zapateo: ?

zapateo: I would rephrase this sentence, because it implies that an ordering group itself can have one or more delivery endpoints, whereas in fact it is the events associated with a specific ordering group that can have one or more delivery endpoints.

> verifies that its ordering group uses a single delivery endpoint

Does what I'm saying make sense to you, or am I misunderstanding something?

zapateo: It refers to "nil entries" here, but that strikes me as a more technical, implementation-level concept; from the caller's perspective, it isn't entirely clear why there would be "nil entries."

What do you think?

connectors/applications.go
```diff
@@ -234,6 +234,9 @@ func (r FailureReason) String() string {
 // MaxOrderingGroupLen is the maximum length of an event ordering group.
 const MaxOrderingGroupLen = 16
 
+// MaxDeliveryEndpointLen is the maximum length of a delivery endpoint.
+const MaxDeliveryEndpointLen = 25
```
zapateo: As discussed verbally, this value should be aligned with the maximum length of the ordering groups; therefore, it should be 16.

core/internal/connections/application.go
```diff
@@ -168,6 +179,14 @@ func (app *Application) EventTypes(ctx context.Context) ([]*EventType, error) {
 				return nil, fmt.Errorf(
 					"connector %s returned multiple event types with the same ID (%s)", app.connector, eventType.ID)
 			}
+			if next == nil || next.OrderingGroup != eventType.OrderingGroup {
```
zapateo: *Note that the AI pointed this out to me, so evaluate it with caution.*

---

`validateEventType` runs on `eventTypes[i]` only, while the pairwise comparison reaches `eventTypes[i+1:]` before those entries have been validated themselves. So for

```go
[]*EventType{
    {ID: "createContact", OrderingGroup: "contacts"},
    {ID: "updateContact", OrderingGroup: "contacts", DeliveryEndpoint: "contact-events"},
}
```

the first iteration returns

```
connector test returned different delivery endpoints for ordering group "contacts"
```

and the real problem — `"contact-events"` is not a valid property name — is never reported. The connector author is pointed at the wrong event type.

The ordering group doesn't have this issue: an invalid group on a later entry makes `next.OrderingGroup != eventType.OrderingGroup` true, the comparison is skipped, and the outer loop reaches it and reports it properly. It's specific to the endpoint.

Validating every entry in a first pass and doing the pairwise checks in a second one would give the specific message in both cases.

core/internal/connections/application.go
```diff
@@ -112,8 +112,9 @@ func (app *Application) Connector() string {
 
 // EventType returns the application's event type with the specified ID. The
 // returned event type is owned by the connector and must not be modified.
-// It validates only the matching event type and rejects duplicates of its ID;
-// nil entries and other event types are ignored.
+// It validates the matching event type, rejects duplicates of its ID, and
+// verifies that its ordering group uses a single delivery endpoint. Nil entries
+// and event types in other ordering groups are ignored.
```
zapateo: *Note that the AI pointed this out to me, so evaluate it with caution.*

> It validates the matching event type, rejects duplicates of its ID, and verifies that its ordering group uses a single delivery endpoint. Nil entries and event types in other ordering groups are ignored.

Saying that event types in *other* ordering groups are ignored suggests that those in the same group are not, but they are still unvalidated: only their `DeliveryEndpoint` is compared, their own ID and ordering group are never checked. The old wording ("nil entries and other event types are ignored") didn't raise that expectation.

Maybe something closer to: only the matching event type is validated; the others are read solely to compare the delivery endpoint of its ordering group.

zapateo (changes requested):

core/connection.go
gazerro: See PR https://github.com/krenalis/krenalis.com/pull/67.

core/internal/connections/application.go
gazerro: Changed in commit [Clarify Application.EventType comment](https://github.com/krenalis/krenalis/pull/2401/changes/3da6a1889f3a6c1de4761a972b4d666035c9bdca).

gazerro: Changed in commit [Clarify Application.EventType comment](https://github.com/krenalis/krenalis/pull/2401/changes/3da6a1889f3a6c1de4761a972b4d666035c9bdca).

connectors/applications.go
gazerro: Changed in commit [Change maximum delivery endpoint length from 25 to 16](https://github.com/krenalis/krenalis/pull/2401/changes/691253f2d6269e5fa405420b1b8fa015e87d2947).

core/internal/connections/application.go
gazerro: I clarified in the error message that the `DeliveryEndpoint` fields are different: [Clarify error massage](https://github.com/krenalis/krenalis/pull/2401/changes/3c1ffad4981c7412cf3774aed36c69e9d0ec3cb3).

core/internal/connections/application.go
gazerro: Changed in commit [Clarify Application.EventType comment](https://github.com/krenalis/krenalis/pull/2401/changes/3da6a1889f3a6c1de4761a972b4d666035c9bdca).

connectors/skills/create-application/references/events.md
gazerro: See PR https://github.com/krenalis/krenalis.com/pull/67.

connectors/applications.go
zapateo: Remember to update the commit message, which still says 25.

core/internal/connections/application_test.go
```diff
@@ -350,6 +355,16 @@ func TestValidateEventType(t *testing.T) {
...
 			err:       `connector test returned an invalid ordering group ("aaaaaaaaaaaaaaaaa")`,
 		},
+		{
+			name:      "invalid delivery endpoint",
+			eventType: &EventType{ID: "contact", OrderingGroup: "contacts", DeliveryEndpoint: "contact-events"},
+			err:       `connector test returned an invalid delivery endpoint ("contact-events")`,
+		},
+		{
+			name:      "long delivery endpoint",
+			eventType: &EventType{ID: "contact", OrderingGroup: "contacts", DeliveryEndpoint: strings.Repeat("a", 26)},
```
zapateo: *Note that the AI pointed this out to me, so evaluate it with caution.*

This looks like a leftover from when the limit was 25. With a limit of 16, a 26-character value would still fail if the limit went up to 20, so the case no longer pins the boundary.

The ordering group tests have a valid 16-character case and an invalid 17-character one (the "16 character ordering group" and "long ordering group" rows). Doing the same here would make the two sets of cases match.

core/internal/initdb/upgrade_test.go
```diff
@@ -276,14 +277,24 @@ func TestUpgradePipelineOrderingGroup(t *testing.T) {
...
 					ALTER TABLE pipelines DROP COLUMN ordering_group;
 					ALTER TABLE pipelines ALTER COLUMN event_type TYPE varchar(100)`)
 				if err != nil {
 					t.Fatal(err)
 				}
 			}
+			if name == "delivery-appended" {
+				_, err = database.Exec(t.Context(), `
+					ALTER TABLE pipelines DROP COLUMN delivery_endpoint;
+					ALTER TABLE pipelines ADD COLUMN delivery_endpoint varchar(25);
```
zapateo: *Note that the AI pointed this out to me, so evaluate it with caution.*

No released schema ever had `delivery_endpoint varchar(25)`. It only existed on this branch before the limit changed to 16. The fixture should simulate a state a real database can be in, so I'd use `varchar(16)` here.

core/internal/connections/application_test.go
zapateo: *Note that the AI pointed this out to me, so evaluate it with caution.*

This test now also checks that event types in the same ordering group use the same delivery endpoint, so I'd mention it here too, as you did for `TestApplicationEventType`. For example:

```go
// TestApplicationEventTypes verifies independent IDs, mandatory ordering
// groups, and consistent delivery endpoints.
```

zapateo (changes requested):

connectors/applications.go
gazerro: Done.

core/internal/connections/application_test.go
gazerro: Fixed in commit [Fix TestUpgradePipelineOrderingGroup test](https://github.com/krenalis/krenalis/pull/2401/changes/691d5438afea394938415c81678a0820e7eead0b).

core/internal/initdb/upgrade_test.go
gazerro: Fixed in commit [Fix test](https://github.com/krenalis/krenalis/pull/2401/changes/bc7657e99cf34f15fcfcee9cb5b291e098855afb).

core/internal/connections/application_test.go
gazerro: Done in commit [Clarify test name and comment](https://github.com/krenalis/krenalis/pull/2401/changes/09dfd096885b907dd4bba3b1471f862a9bcc320a).

core/internal/connections/application_test.go
zapateo: I assume you meant to refer to commit https://github.com/krenalis/krenalis/pull/2401/changes/c49e69874a6275d13c41b7d23d8453dd983442c5. Anyway, it's all good; I'm marking it as resolved.

zapateo (approved):

PR #2402, opened 2026-08-04
core/internal/collector: reset the event decoder type between requests

gazerro:
```
core/internal/collector: reset the event decoder type between requests (#2402)

A decoder retained the endpoint-specific event type after being reset.
When reused, it could therefore interpret a generic "/events" request
without a type as using the previous endpoint's type.

Clear the type in Reset and add regression coverage for decoder reuse.
```

### DIFF
--- a/core/internal/collector/decoder.go
+++ b/core/internal/collector/decoder.go
@@ -207,6 +207,7 @@ func (d *decoder) Reset(r *http.Request) error {
 	d.writeKey = ""
 	d.connectionId = ""
 	d.context = nil
+	d.typ = ""
 
 	path, _ := strings.CutPrefix(r.URL.Path, "/events")
 	switch path {
--- a/core/internal/collector/decoder_test.go
+++ b/core/internal/collector/decoder_test.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"net/netip"
 	"net/url"
 	"reflect"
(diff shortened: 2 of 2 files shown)


core/internal/collector/decoder.go
```diff
@@ -207,6 +207,7 @@ func (d *decoder) Reset(r *http.Request) error {
 	d.writeKey = ""
 	d.connectionId = ""
 	d.context = nil
+	d.typ = ""
```
zapateo, line 210: We could consider moving all the reset assignments to the beginning of the `Reset` function (they're currently split into two blocks), so the decoder would be completely reset every time, even when it exits on error; I mean something like this:

```diff
diff --git a/core/internal/collector/decoder.go b/core/internal/collector/decoder.go
index deb7f0b78..cd382b122 100644
--- a/core/internal/collector/decoder.go
+++ b/core/internal/collector/decoder.go
@@ -145,6 +145,18 @@ func (d *decoder) Events(connectionId string, fallbackToRequestIP bool) iter.Seq
 //   - a badRequestError: if the request's body is not valid.
 func (d *decoder) Reset(r *http.Request) error {
 
+	d.batch = nil
+	d.receivedAt = time.Now().UTC()
+	d.remoteAddr.ip = netip.Addr{}
+	d.remoteAddr.identifiable = ""
+	d.remoteAddr.partiallyAnonymous = ""
+	d.remoteAddr.stronglyAnonymous = ""
+	d.sentAt = time.Time{}
+	d.writeKey = ""
+	d.connectionId = ""
+	d.context = nil
+	d.typ = ""
+
 	if r.Method != "POST" {
 		return errMethodNotAllowed
 	}
@@ -166,10 +178,6 @@ func (d *decoder) Reset(r *http.Request) error {
 		}
 	}
 
-	d.batch = nil
-	d.receivedAt = time.Now().UTC()
-	d.remoteAddr.ip = netip.Addr{}
-
 	// If the 'X-Forwarded-For' header is present, use it to determine
 	// the client's IP address. Also accept non-standard formats such as
 	// bracketed IPv6 addresses and addresses with a ":port" suffix.
@@ -203,12 +211,6 @@ func (d *decoder) Reset(r *http.Request) error {
 		}
 	}
 
-	d.sentAt = time.Time{}
-	d.writeKey = ""
-	d.connectionId = ""
-	d.context = nil
-	d.typ = ""
-
 	path, _ := strings.CutPrefix(r.URL.Path, "/events")
 	switch path {
 	case "":

```

zapateo (changes requested):

core/internal/collector/decoder.go
gazerro: Done in commit [Reset decoder state before request validation](https://github.com/krenalis/krenalis/pull/2402/changes/7cf0a68e57a7baa79408e1e1ffc3ebfd11b02f3e).

See also [Format TestDecoderResetClearsType](https://github.com/krenalis/krenalis/pull/2402/changes/c32c59d55665e711229b1da5218edc8c264e6277).

zapateo (changes requested):
Please fix the conflict with `main`.

gazerro:
> Please fix the conflict with `main`.

Done.

zapateo (approved):

PR #2406, opened 2026-08-05
core: track event acknowledgments across destinations

gazerro:
```
core: track acknowledgments per event destination (#2406)

A single event can be delivered through multiple destination pipelines.
In the existing implementation, those pipelines share a single
acknowledgment, so the first one to finish can acknowledge the NATS
message while the others are still processing their copies of the event.

Missing or closed pipelines also cause related issues:

 * events for those pipelines can be dropped without invoking the shared
   acknowledgment;
 * closing a pipeline can leave gaps in the sender's per-user sequence,
   preventing later events for that user from progressing.

This change gives each event destination its own idempotent
acknowledgment, coordinated through shared state associated with the
original NATS message.

As a result:

 * acknowledging one destination completes only that branch;
 * the NATS message is acknowledged only when all destinations are
   complete;
 * destinations whose pipelines are missing or closed are marked
   complete;
 * when a pipeline closes, its queued sender events are discarded before
   their destination branches are marked complete, allowing per-user
   ordering to continue.

A separate planned change will periodically send InProgress
acknowledgments to JetStream while a message is being processed. Each
acknowledgment resets the message's AckWait redelivery timer. The
completion paths in this change ensure that missing or closed pipelines
cannot cause these acknowledgments to continue indefinitely.

This change does not limit event processing time or alter delivery retry
behavior. Suspending unavailable destinations and moving long-lived
delivery work out of memory will be addressed separately.
```

### DIFF
--- a/core/internal/streams/nats/nats.go
+++ b/core/internal/streams/nats/nats.go
@@ -554,7 +554,7 @@ func (s *stream) Consume(topic string, size int) streams.Consumer {
 
 // ack coordinates the destination acknowledgments of a NATS message.
 type ack struct {
-	mu        sync.Mutex // protects remaining and the done field of each destination acknowledgment.
+	mu        sync.Mutex // protects 'remaining' and every destinationAck.done field.
 	msg       jetstream.Msg
 	remaining int // number of destinations that have not acknowledged the message; protected by mu.
 }
--- a/core/internal/streams/nats/nats.go
+++ b/core/internal/streams/nats/nats.go
@@ -511,18 +511,19 @@ func (s *stream) Consume(topic string, size int) streams.Consumer {
 						err = fmt.Errorf("invalid event data: %s", err)
 						return
 					}
+					var destinations []string
 					if header := msg.Headers(); header != nil {
-						if destinations, ok := header["destinations"]; ok {
-							for _, d := range destinations {
-								if !isValidDestination(d) {
-									err = fmt.Errorf("invalid event destination: %q", d)
+						if ids, ok := header["destinations"]; ok {
+							for _, id := range ids {
+								if !isValidDestination(id) {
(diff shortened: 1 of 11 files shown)


core/internal/streams/nats/nats.go
```diff
@@ -551,15 +552,37 @@ func (s *stream) Consume(topic string, size int) streams.Consumer {
 	return consumer
 }
 
-// ack acknowledges a NATS event.
+// ack coordinates the destination acknowledgments of a NATS message.
 type ack struct {
-	msg jetstream.Msg
+	mu        sync.Mutex // protects remaining and the done field of each destination acknowledgment.
```
zapateo: I think it's not so clear that this commit refers to the `destinationAck.done` field when talking about "_done field_".

Moreover, since "_remaining_" and "_done_" are words commonly used in English, I would put them in quotes to avoid ambiguity.

core/internal/streams/nats/nats.go
zapateo, line 1: The commit message refers to _acknowledgment heartbeats_, but it's not clear to me exactly what mechanism this refers to, nor whether the sentence refers to the current state on `main` or whether it refers to a hypothetical or future scenario.

zapateo (changes requested):

core/internal/streams/nats/nats.go
gazerro: Fixed in commit [Clarify comment](https://github.com/krenalis/krenalis/pull/2406/changes/0ecf053620d7b4251e58f6a91e1a257094e9c2c5).

core/internal/streams/nats/nats.go
gazerro: I reviewed and improved the commit message.

zapateo (approved):

PR #2407, opened 2026-08-06
Add consent management to user pipelines

retini:


core/pipeline.go
```diff
@@ -925,6 +929,20 @@ func (this *Pipeline) Update(ctx context.Context, pipeline PipelineToSet) error
 				}
 			}
 		}
+		// Check that the required consent purposes exist.
+		if len(n.RequiredConsents.Purposes) > 0 {
+			var missing string
+			err := tx.QueryRow(ctx, "SELECT purpose\n"+
+				"FROM UNNEST($1::varchar[]) AS purpose\n"+
+				"WHERE NOT EXISTS (SELECT FROM consent_purposes AS cp WHERE cp.id = purpose AND cp.workspace = $2)\n"+
```
gazerro: I'd use `SELECT 1 FROM`, but it's just a minor stylistic detail that makes the intent of the query more explicit without changing its behavior.

core/pipeline.go
```diff
@@ -925,6 +929,20 @@ func (this *Pipeline) Update(ctx context.Context, pipeline PipelineToSet) error
...
+				"LIMIT 1", n.RequiredConsents.Purposes, c.Workspace().ID).Scan(&missing)
+			if err != nil && err != sql.ErrNoRows {
```
gazerro: I think using `missing != ""` instead of `err == nil` as the `if` condition makes the intent clearer. If you agree, please use it.

```go
if err != nil && err != sql.ErrNoRows {
     return nil, err
}
if missing != "" {
     return nil, errors.Unprocessable(ConsentPurposeNotExist, "consent purpose %s does not exist", missing)
}
```

gazerro (commented):
Everything looks great. I just have two small comments on a section of code that we ended up not needing to review last time.

core/pipeline.go
retini: Done in commit [`Review the checks done on required consent purposes before writing pipelines`](https://github.com/krenalis/krenalis/pull/2407/changes/d38f0fca70b91b6143c7936212573b2b190e5fca).

Additionally, as discussed, the commit also:

  - moves the check in `Pipeline.Update` before the first write of the transaction, so that it now runs ahead of the `INSERT INTO discontinued_functions` instead of after it;
  - adds the same check to `Connection.CreatePipeline`, as the first statement of the transaction, where it was missing altogether: creation validated the required purposes against the state only.

core/pipeline.go
retini: Done in [`Detect missing consent purposes by the scanned identifier instead of the query error`](https://github.com/krenalis/krenalis/pull/2407/changes/d057d491f82af2b3659f50e10842722eb76a6f8a)

gazerro (changes requested):
Please check the document that I shared with you separately.

retini:
> Please check the document that I shared with you separately.

@gazerro, I went through the document and addressed the feedback as follows.

### Review findings

- 1 and 5 — no changes were made. Consent purposes have not been released and are not currently used by any production instance, so no released database can contain the existing purpose paths or unresolved purpose references described by these findings.
- 2 — addressed in 35b0d4156. The migration starts from the last released six-step layout, since the intermediate consent step currently on main has not been released.
- 3 — addressed in 35b7e1a6a.
- 4 — addressed in the same commit of #3, 35b7e1a6a.
- 6 — I left this for you to review and address, since it concerns the Admin profile schema UI you worked on.

### Repository conventions

- 1 — intentionally unchanged. As discussed, methods such as `ImportProfileConsentFailed` are convenience methods whose purpose is to simplify callers by delegating to the generic exported metric operation. We agreed that allowing an exported method to call another exported method is reasonable in this specific case. Please consider whether this exception should be documented in `AGENTS.md`.
- 2 — addressed in 295677a9d.
- 3 — addressed in 67ec28f95.
- 4 — addressed in 6ae791345.

gazerro:
I would change the UI on the pipeline screen as follows:

image

I would also use a switch instead of a checkbox, and I would keep the all/any selector enabled but make it read-only rather than disabled.

retini:
> I would change the UI on the pipeline screen as follows:
> 
> image
> I would also use a switch instead of a checkbox, and I would keep the all/any selector enabled but make it read-only rather than disabled.

@gazerro, I updated the UI as you suggested, in commit [`Redesign the pipeline consent requirements section`](https://github.com/krenalis/krenalis/pull/2407/changes/d3fc54dd4570c7954e0738f7c4663a08339b7e18).

core/pipeline_validate.go
```diff
@@ -165,20 +172,24 @@ func validatePipelineToSet(pipeline PipelineToSet, v validationState) error {
...
+		for i, id := range pipeline.RequiredConsents.Purposes {
+			if !IsValidID(id) {
+				return errors.BadRequest("identifier %q is not a valid consent purpose identifier", id)
+			}
+			if slices.Contains(pipeline.RequiredConsents.Purposes[i+1:], id) {
+				return errors.BadRequest("required consent purpose %s is duplicated", id)
 			}
-			if slices.Contains(pipeline.RequiredConsents.Purposes[i+1:], code) {
-				return errors.BadRequest("required consent purpose %q is duplicated", code)
+			if !v.knownConsentPurposeIDs[id] {
```
gazerro: This validation is already performed within the transaction, so doing it here as well only optimizes an error case.

I also don't think this is the right place for this validation, since this function doesn't perform the other validations that are later handled within the transaction either.

I would remove this validation altogether.

core/pipeline_validate.go
```diff
@@ -165,20 +172,24 @@ func validatePipelineToSet(pipeline PipelineToSet, v validationState) error {
 		}
 	}
 	// Validate the required consents.
-	requiredConsentsAllowed := dispatchEventsToAplications || importEventsIntoWarehouse || importUserIdentitiesFromEvents
+	requiredConsentsAllowed := dispatchEventsToAplications || importEventsIntoWarehouse ||
```
gazerro: Isn't `requiredConsentsAllowed` always `true` now that consent checks are performed in all pipelines?

core/internal/state/state.go
```diff
@@ -1335,8 +1358,64 @@ func (account *Account) Connector() *Connector {
...
+	ProfileConsentLocation *ProfileConsentLocation
+	// eventPropertyPaths and profilePropertyPath are the paths actually read,
+	// resolved by resolvePropertyPaths.
+	eventPropertyPaths  [][]string // never nil
+	profilePropertyPath []string
+}
+
+// NewConsentPurpose returns a new consent purpose with the paths of its
+// properties resolved.
+func NewConsentPurpose(purpose ConsentPurpose) *ConsentPurpose {
```
gazerro: There aren't any other `New[...]` functions like this in the state. Inline it where it's used.

PR #2408, opened 2026-08-06
core/internal/collector/sender: separate ordering queues by group

gazerro:
```
core/internal/collector/sender: separate ordering queues by group

Previous changes stored ordering groups and delivery endpoints on
destination pipelines without changing event delivery. The sender still
uses a single logical ordering queue for each anonymous ID, so all event
types for a user remain coupled regardless of their ordering group.

Pass each pipeline's ordering group to Sender.CreateEvent and identify
each logical ordering by both ordering group and anonymous ID. Pipelines
in the same group share a creation sequence, while different groups for
the same user can be processed independently.

Keep Events.SameUser bound to the anonymous ID rather than to a specific
ordering. This allows connectors to include events from all of a user's
ordering groups in a single iteration. Release each ordering
independently after its final event has been sent or discarded.

This commit changes only the sender's in-memory ordering. It retains one
sender and one physical queue per connection, and leaves batching and
connector requests unchanged. Acknowledgment aggregation and NATS
routing by ordering group, delivery endpoint, and shard are
intentionally deferred to follow-up changes.
```


core/internal/initdb/upgrade.go
zapateo: This change shouldn't be part of this PR, or am I mistaken?

Same for `core/internal/initdb/upgrade_test.go`.

core/internal/collector/sender/sender_randomized_test.go
zapateo: *Note that the AI pointed this out to me, so evaluate it with caution.*

The randomized test is the one that really stresses the postpone / release / `available` bookkeeping. But it always passes `"events"`, so it never has more than one ordering per user, and that's exactly what's new in this change. What about picking the group at random from a small set, and checking the delivery order per (group, anonymous ID)?

core/internal/collector/sender/user.go
zapateo: The file name should be kept the same so that Git do not consider a deleted file and an added file.

We can then rename the file on the `main` branch; this ensures the commit resulting from this branch clearly displays the diff, while the renaming commit performs one single, specific action.

zapateo (changes requested):

gazerro:
See commit [Rename ordering.go to user.go](https://github.com/krenalis/krenalis/pull/2408/changes/e81fb1a5d75b7b21f6c52e71ade67a0bcf3d6ba6).

core/internal/collector/sender/sender.go
```diff
@@ -226,24 +226,25 @@ func (s *Sender) Close(ctx context.Context) {
...
+// group, schema, attributes, and acknowledgment.
 //
 // The caller must pass the returned event to SendEvent, optionally after
 // setting Properties, or to DiscardEvent if the event should not be sent.
-func (s *Sender) CreateEvent(pipeline, typ string, schema types.Type, attributes map[string]any, ack streams.Ack) *Event {
+func (s *Sender) CreateEvent(pipeline, typ, orderingGroup string, schema types.Type, attributes map[string]any, ack streams.Ack) *Event {
 	anonymousID, ok := attributes["anonymousId"].(string)
 	if !ok {
 		panic("CreateEvent called with an event missing anonymousId")
 	}
```
zapateo: Here, there is a check to ensure the anonymous ID exists and is a string. Since the change in this branch gives a specific, significant meaning to an empty anonymous ID:

https://github.com/krenalis/krenalis/blob/9e4a1d70d88a0e781b19b3558c67f7da89bc291b/core/internal/collector/sender/sender.go#L560

Would it make sense to add to `CreateEvent` a check ensuring the anonymous ID isn't an empty string and panic if it is? The decoder currently ensures that it is not an empty string, but a panic might reinforce that requirement (both to avoid unexpected behavior in case of changes and for documentation purposes).

What do you think?

core/internal/collector/sender/sender_test.go
```diff
@@ -286,6 +286,134 @@ func Test_Sender_DiscardedOutOfOrderEvent(t *testing.T) {
 	})
 }
 
+// Test_Sender_OrderingGroupPreservesOrder verifies that events for the same
+// user and ordering group are delivered in creation order even when associated
+// with different pipelines.
+func Test_Sender_OrderingGroupPreservesOrder(t *testing.T) {
```
zapateo: Please review the blank lines in this function, as they do not appear to comply with the AGENTS.md file.

Same for other functions added in this file.

core/internal/collector/sender/sender.go
zapateo: The commit message states:

> Pipelines in the same group share a creation sequence

Isn't the "creation sequence" aspect a bit too focused on implementation details? Wouldn't it be better to describe the actual behavior observed in practice?

But it's up to you—this is just a suggestion.

zapateo (changes requested):

core/internal/initdb/upgrade.go
gazerro: Fixed in commit [Revert upgrade changes](https://github.com/krenalis/krenalis/pull/2408/changes/4bef27eeec2182613f2636745dfeadbfa59db556).

core/internal/collector/sender/user.go
gazerro: Done in commit [Rename ordering.go to user.go](https://github.com/krenalis/krenalis/pull/2408/changes/e81fb1a5d75b7b21f6c52e71ade67a0bcf3d6ba6).

core/internal/collector/sender/sender.go
gazerro: As discussed, I used `nil` instead of an empty string to represent no user in `sameUser`.

See commit [Use nil instead of an empty string for no user in sameUser](https://github.com/krenalis/krenalis/pull/2408/changes/c631b9717f1d598cd62fa5c88a9365a5e8f26537).

core/internal/collector/sender/sender_test.go
gazerro: Done in commit [Fix test formatting](https://github.com/krenalis/krenalis/pull/2408/changes/69485b41726975ad4d7646dc6a49097a84878c19).

core/internal/collector/sender/sender.go
gazerro: Done.

PR #2410, opened 2026-08-09
core, warehouses: add identity counts to profiles

gazerro:
```
core, warehouses: add identity counts to profiles (#2410)

Add _anonymous_count and _recognized_count to profile tables and
populate them during Identity Resolution.

The counts represent the number of distinct anonymous and non-anonymous
identities that are part of each profile, without counting the same
identity multiple times when processed by multiple pipelines.

Existing warehouse schemas are not migrated by this commit.
```

### DIFF
--- a/core/internal/datastore/warehouses_identityresolution_test.go
+++ b/core/internal/datastore/warehouses_identityresolution_test.go
@@ -56,6 +56,11 @@ type identity struct {
 	attributes   map[string]any
 }
 
+type identityCounts struct {
+	anonymous  int
+	recognized int
+}
+
 const (
 	identityResolutionConnection1 = "6NpT4zB8QaR2"
 	identityResolutionConnection2 = "8QaT3mN7KxP5"
@@ -89,7 +94,14 @@ func TestWarehousesIdentityResolution(t *testing.T) {
 		primarySources   map[string]string
 		identities       []identity
 		expectedProfiles []map[string]any
+		expectedCounts   []identityCounts
 	}{
+		{
+			name:           "No identities",
+			identifiers:    []string{},
+			identities:     []identity{},
+			expectedCounts: []identityCounts{},
+		},
 		{
 			name:        "One identity, no identifiers",
 			identifiers: []string{},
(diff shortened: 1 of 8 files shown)


warehouses/postgresql/identityresolution.go
```diff
@@ -91,6 +91,13 @@ func (warehouse *PostgreSQL) ResolveIdentities(ctx context.Context, opID string,
 		if err != nil {
 			return fmt.Errorf("cannot create profiles table (with name %s): %s", quoteIdent(newProfilesName), err)
 		}
+		// Add the identity count columns if they are not already present.
+		_, err = tx.Exec(ctx, `ALTER TABLE `+quoteIdent(newProfilesName)+
```
zapateo: I would clarify that this code isn't meant to remain in production, but is intended solely for the update process. I might add a TODO note.

This also applies to Snowflake.

core/internal/datastore/warehouses_identityresolution_test.go
```diff
@@ -56,6 +56,11 @@ type identity struct {
 	attributes   map[string]any
 }
 
+type identityCounts struct {
```
zapateo: Here, I would clarify that this isn't a count of "identities" (as we currently understand them in the software), but rather a count of "logical identities." I find this distinction important in a specific context like a test, because introducing misunderstandings at this level can lead to confusion—both for us humans and for AIs.

The terminology in the other tests modified by this PR should also be reviewed slightly, following the same criterion.

warehouses/postgresql/identityresolution.go
```diff
@@ -181,6 +188,8 @@ func (warehouse *PostgreSQL) ResolveIdentities(ctx context.Context, opID string,
 	}
 	// Write the "_identities" column.
 	mergeProfiles.WriteString(`ARRAY_AGG(DISTINCT "_pk"), `)
+	mergeProfiles.WriteString(`COUNT(DISTINCT ("_connection", "_identity_id")) FILTER (WHERE "_is_anonymous"), `)
```
zapateo: Here, I would add a comment clarifying that the pipeline is not included in the `DISTINCT` operation because the goal is not to count identities as currently defined in the software (the "row identities"), but rather to count "logical identities" that are unified at the pipeline level.

This also applies to Snowflake.

warehouses/postgresql/initrepair.go
zapateo: (I'm commenting here because I don't know where else to put the comment.)

The commit message already correctly explains the type of identity being counted (i.e., the identities grouped by pipeline); I would simply add the term "logical identity"—perhaps even in the title—so that in the future, it will be easier for us and for AIs to understand that a distinction has been introduced.

zapateo (changes requested):

warehouses/postgresql/identityresolution.go
gazerro: Done in commit [Clarify that logical identity columns are temporary](https://github.com/krenalis/krenalis/pull/2410/changes/d446895e50f0d93d82ba8ad50893edc74b2d0828).

core/internal/datastore/warehouses_identityresolution_test.go
gazerro: Done in commit [Clarify raw identities vs. logical identities](https://github.com/krenalis/krenalis/pull/2410/changes/8d3110f496661d8e19e847b766db1ad91600c2f0).

warehouses/postgresql/identityresolution.go
gazerro: Fixed in commit [Add missing comments to identity resolution queries](https://github.com/krenalis/krenalis/pull/2410/changes/a2c8e701da5317b5792ae6031a991d3cac1d1b95).

warehouses/postgresql/initrepair.go
gazerro: Done.

zapateo (approved):

PR #2414, opened 2026-08-13
core/internal/state: prevent panic when loading source-only connectors

gazerro:
```
core/internal/state: prevent panic when loading source-only connectors (#2414)

If a connector is source-only and not a destination, loading the state
causes a panic when accessing connector.AsDestination.SendingMode
because connector.AsDestination is nil.
```

### DIFF
--- a/core/internal/state/load.go
+++ b/core/internal/state/load.go
@@ -72,13 +72,15 @@ func (state *State) load(ctx context.Context, oauthCredentials map[string]*OAuth
 			if connector.Terms.UserID != "" {
 				c.Terms.UserID = connector.Terms.UserID
 			}
-			switch connector.AsDestination.SendingMode {
-			case connectors.Client:
-				c.SendingMode = new(Client)
-			case connectors.Server:
-				c.SendingMode = new(Server)
-			case connectors.ClientAndServer:
-				c.SendingMode = new(ClientAndServer)
+			if connector.AsDestination != nil {
+				switch connector.AsDestination.SendingMode {
+				case connectors.Client:
+					c.SendingMode = new(Client)
+				case connectors.Server:
+					c.SendingMode = new(Server)
+				case connectors.ClientAndServer:
+					c.SendingMode = new(ClientAndServer)
+				}
 			}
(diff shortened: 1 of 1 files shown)


core/internal/state/load.go
```diff
@@ -72,13 +72,15 @@ func (state *State) load(ctx context.Context, oauthCredentials map[string]*OAuth
...
 				c.Terms.UserID = connector.Terms.UserID
 			}
-			switch connector.AsDestination.SendingMode {
-			case connectors.Client:
-				c.SendingMode = new(Client)
-			case connectors.Server:
-				c.SendingMode = new(Server)
-			case connectors.ClientAndServer:
-				c.SendingMode = new(ClientAndServer)
+			if connector.AsDestination != nil {
```
zapateo: Some lines above we have:

https://github.com/krenalis/krenalis/blob/7f1cb6b8d4e8f1f350e71a10c25bc90bdbeaa07f/core/internal/state/load.go#L60-L65

So we might consider, instead of introducing an new "if" with this PR, simply moving this SendingMode code into the if statement we already have above, getting something like:

```go
if asDest := connector.AsDestination; asDest != nil {
	c.DestinationTargets = ConnectorTargets(asDest.Targets)
	c.HasDestinationSettings = asDest.HasSettings
	c.Documentation.Destination.Summary = asDest.Documentation.Summary
	c.Documentation.Destination.Overview = asDest.Documentation.Overview
	switch asDest.SendingMode {
	case connectors.Client:
		c.SendingMode = new(Client)
	case connectors.Server:
		c.SendingMode = new(Server)
	case connectors.ClientAndServer:
		c.SendingMode = new(ClientAndServer)
	}
}
```

Which, in my opinion, is more compact and logical. What do you think? If you say it's not worth it, I'll simply approve the change as it is now.

zapateo (changes requested):

core/internal/state/load.go
gazerro: The fields are initialized in an order that mirrors their declaration in the struct. I'd prefer to keep the `SendingMode` assignment in that order as well, just to preserve that consistency and keep the code a bit clearer.

zapateo (approved):

PR #2415, opened 2026-08-13
core/internal/connections: restore event iterator state after preview

gazerro:
```
core/internal/connections: restore event iterator state after preview (#2415)

Peek, Discard, and Postpone cannot be called after an iteration has
completed; doing so correctly causes a panic.

However, the single-event iterator used by PreviewSendEvent did not
reset its iterating field after All or SameUser completed. As a result,
Peek returned no event, and Discard and Postpone did not panic.

Reset the iterator state and add tests covering the single-event
iterator before, during, and after All and SameUser iterations.
```

### DIFF
--- a/core/internal/connections/application.go
+++ b/core/internal/connections/application.go
@@ -401,6 +401,7 @@ func (iter *singleEventIterator) All() iter.Seq[*connectors.Event] {
 	return func(yield func(event *connectors.Event) bool) {
 		iter.iterating = true
 		yield(iter.event)
+		iter.iterating = false
 	}
 }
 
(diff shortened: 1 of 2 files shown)


core/internal/connections/application.go
zapateo: It appears this change also fixes a `Discard` issue, in addition to `Peek`, as discussed verbally.

So we can consider:

1. Changing the commit message to include `Discard`
2. Adding a test for `Discard`, if possible

zapateo (changes requested):

core/internal/connections/application.go
gazerro: Done in commit [Test Discard and Postpone panic after event preview](https://github.com/krenalis/krenalis/pull/2415/changes/016729aa4dfdfbd435083da4ee312e5a9415b85d). Note that I also added `Postpone`, as it is affected in the same way.

zapateo (changes requested):
Please resolve the merge conflicts.

gazerro:
> Please resolve the merge conflicts.

Done.

zapateo (approved):

PR #2416, opened 2026-08-13
core/internal/collector/sender: test `Peek` lifecycle

gazerro:
```
core/internal/collector/sender: test `Peek` lifecycle (#2416)

Add coverage for Events.Peek before, during, and after All and SameUser
iterations.

Verify that repeated calls before iteration do not consume the event and
that Peek panics when called after the iteration completes.
```

### DIFF
--- a/core/internal/collector/sender/sender_test.go
+++ b/core/internal/collector/sender/sender_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"iter"
 	"math"
 	"slices"
 	"testing"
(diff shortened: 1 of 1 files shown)


core/internal/collector/sender/sender_test.go
```diff
@@ -79,6 +80,75 @@ func Test_newStoppedTimer(t *testing.T) {
...
+						if got, ok := events.Peek(); ok || got != nil {
+							t.Fatalf("Peek during iteration: expected nil and false, got %p and %t", got, ok)
+						}
+					}
+					if yielded != 1 {
+						t.Fatalf("expected one event, got %d", yielded)
+					}
+					func() {
+						defer func() {
+							if recover() == nil {
```
zapateo: This check (`recover() == nil`) seems a bit too general to me, and if there's a panic for other reasons, the test might hide them and pass.

What about making the check more specific? Something like that; see you if it makes sense and is worth.

```go
r := recover()
msg, ok := r.(string)
if !ok || !strings.Contains(msg, "Events.Peek outside of an iteration") {
    t.Errorf("Peek after iteration: unexpected panic %v", r)
}
```

zapateo (changes requested):

core/internal/collector/sender/sender_test.go
```diff
@@ -79,6 +80,75 @@ func Test_newStoppedTimer(t *testing.T) {
...
+		{name: "All", seq: connectors.Events.All},
+		{name: "SameUser", seq: connectors.Events.SameUser},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			synctest.Test(t, func(t *testing.T) {
+				var checked bool
+				app := newTestApplication()
+				app.SendEventsFunc = func(_ context.Context, events connectors.Events) error {
```
zapateo: The `SendEventsFunc` is called in a separate goroutine:

https://github.com/krenalis/krenalis/blob/1ae8e35bc5aa575ca57bd9df03e3fc0aec36c6ba/core/internal/collector/sender/sender.go#L463-L465

and in its body it calls `t.Fatalf`, which in turns calls `FailNow`, whose [documentation](https://pkg.go.dev/testing#T.FailNow) states:

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test.

I'm not sure what kind of repercussions a test like this could have and if and what problems could arise.

core/internal/collector/sender/sender_test.go
gazerro: Fixed in commit [Ensure that the panic is the expected one](https://github.com/krenalis/krenalis/pull/2416/changes/6c82882ed00d05fc11d3db8b1a37b8ec2e803328).

zapateo (changes requested):

core/internal/collector/sender/sender_test.go
gazerro: Fixed in commit [Avoid FailNow from sender test goroutine](https://github.com/krenalis/krenalis/pull/2416/changes/fbf12b2a7617b57fc57f69d64b0d9c30b75370b6).

zapateo (approved):

PR #2421, opened 2026-08-13
core/testconnector: initialize `SameUser` from anonymous ID

gazerro:
```
core/testconnector: initialize `SameUser` from anonymous ID (#2421)

Use the first event's anonymous ID to initialize SameUser's selected
user, matching subsequent comparisons and sender behavior.

Add a regression test with intentionally different user and anonymous
IDs.
```

### DIFF
--- a/core/testconnector/iterator_test.go
+++ b/core/testconnector/iterator_test.go
@@ -1,0 +1,39 @@
+// Copyright 2026 Open2b. All rights reserved.
+// Use of this source code is governed by an Elastic License 2.0
+// that can be found in the LICENSE file.
+
+package testconnector
+
+import (
+	"slices"
+	"testing"
+
+	"github.com/krenalis/krenalis/connectors"
+)
+
+// TestEventsIteratorSameUserUsesAnonymousID verifies that SameUser selects
+// events with the same anonymous ID as the first event, rather than by user ID.
+func TestEventsIteratorSameUserUsesAnonymousID(t *testing.T) {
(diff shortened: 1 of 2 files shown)


core/testconnector/iterator_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/krenalis/krenalis/connectors"
+)
+
+// TestEventsIteratorSameUserUsesAnonymousID verifies that SameUser selects
+// events with the same anonymous ID as the first event, rather than by user ID.
+func TestEventsIteratorSameUserUsesAnonymousID(t *testing.T) {
+	newEvent := func(anonymousID, messageID string) *connectors.Event {
+		return &connectors.Event{Received: ReceivedEvent(map[string]any{
+			"anonymousId": anonymousID,
+			"userId":      "identified-user",
```
zapateo: The commit message states:

> Add a regression test with intentionally different user and anonymous
IDs.

but it looks like the user ID is always the same. In that case, either the test or the commit message should be updated.

zapateo (changes requested):

core/testconnector/iterator_test.go
gazerro: This means that the User ID and Anonymous ID should be different from each other. It does not mean that the User ID should be different for every event. If this is not clear, how would you phrase it?

zapateo: As discussed verbally, we can leave it as it is.

zapateo (approved):

PR #2422, opened 2026-08-13
core/internal/connections/appwriter: test `Records.Peek` lifecycle

gazerro:
```
core/internal/connections/appwriter: test `Records.Peek` lifecycle (#2422)

Cover Peek before, during, and after iterations returned by All and
Same. Verify that it returns the next record without consuming it,
reports when no further records are available, and panics after the
iteration completes.

Also verify that Peek panics after First marks the Records value as
consumed.
```

### DIFF
--- a/core/internal/connections/appwriter/iterator_test.go
+++ b/core/internal/connections/appwriter/iterator_test.go
@@ -6,6 +6,7 @@ package appwriter
 
 import (
 	"iter"
+	"strings"
 	"testing"
 
 	"github.com/krenalis/krenalis/connectors"
@@ -72,8 +73,10 @@ func Test_iterator_Peek(t *testing.T) {
 			}
 
 			defer func() {
-				if recover() == nil {
-					t.Fatal("Peek after iteration: expected panic")
+				r := recover()
+				msg, ok := r.(string)
+				if !ok || !strings.Contains(msg, "Records.Peek outside of an iteration") {
+					t.Errorf("Peek after iteration: unexpected panic %v", r)
 				}
 			}()
 			it.Peek()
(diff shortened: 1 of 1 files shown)


core/internal/connections/appwriter/iterator_test.go
```diff
@@ -0,0 +1,102 @@
...
+					t.Fatalf("Peek at the end of the iteration: expected a zero record and false, got %#v and %t", got, ok)
+				}
+				yielded++
+			}
+			if yielded != len(want) {
+				t.Fatalf("expected %d records, got %d", len(want), yielded)
+			}
+
+			defer func() {
+				if recover() == nil {
```
zapateo: This check only verifies whether a panic occurred, not which one.

This currently hides a problem. If `Peek` does not panic for any reason (it is enough to comment these lines):

https://github.com/krenalis/krenalis/blob/166ac80647f1d4ff06246a5953913d5fd3c56f9a/core/internal/connections/appwriter/iterator.go#L75-L77

The test still passes, because another panic occurs but the test does not notice the difference:

```
runtime error: invalid memory address or nil pointer dereference
```

We could add a check like this:

```go
r := recover()
msg, ok := r.(string)
if !ok || !strings.Contains(msg, "Records.Peek outside of an iteration") {
	t.Errorf("Peek after iteration: unexpected panic %v", r)
}
```

Not sure if the `invalid memory address or nil pointer dereference` is something to fix, or it is just a consequence that `Peek` did not panic and then some code that should not be executed is executed instead.

zapateo (changes requested):

core/internal/connections/appwriter/iterator_test.go
gazerro: Fixed in commit [Ensure that the panic is the expected one](https://github.com/krenalis/krenalis/pull/2422/changes/daa710d219eb59c00a2672d7c016ff6b9162459a).

gazerro: The `invalid memory address or nil pointer dereference` panic is precisely the consequence of `Peek` not panicking.

zapateo (approved):

PR #2427, opened 2026-08-19
all: replace flat filters with compound filter groups

gazerro:
```
all: replace flat filters with compound filter groups (#2427)

The existing filter model applies a single logical operator to all
conditions, making nested logical combinations impossible to express.
It can express:

a AND b AND c

or

a OR b OR c

Introduce compound filter groups that allow conditions to be combined
more flexibly:

a AND (b OR c)

Align filter parsing, validation, persistence, evaluation, and editing
across the backend and Admin console. Migrate existing persisted filters
to the new representation and correct existing inconsistencies in filter
handling.
```

### DIFF
--- a/core/filters.go
+++ b/core/filters.go
@@ -726,7 +726,7 @@ func validateFilter(filter *Filter, schema types.Type, role state.Role, target s
 
 // validateFilterCondition checks the validity of a filter condition and
 // returns its property path.
-func validateFilterCondition(cond *FilterCondition, validation *filterValidation) ([]string, error) {
+func validateFilterCondition(cond *FilterCondition, validation *filterValidation) (string, error) {
 
 	// disallowJSON indicates whether JSON properties are disallowed.
 	disallowJSON := validation.role == state.Destination && validation.target == state.TargetUser
@@ -736,19 +736,19 @@ func validateFilterCondition(cond *FilterCondition, validation *filterValidation
 	disallowEmptyOnObject := validation.role == state.Destination && validation.target == state.TargetUser || validation.target == state.TargetEvent
 
 	if !types.IsValidPropertyPath(cond.Property) {
-		return nil, errors.New("property path is not valid")
+		return "", errors.New("property path is not valid")
 	}
 
 	p, path, err := resolveFilterProperty(validation.properties, cond.Property)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
 	op := cond.Operator
 	kind := p.Type.Kind()
 
 	if disallowJSON && kind == types.JSONKind {
-		return nil, fmt.Errorf("property %q has type json, which is not supported in data warehouse exports", path)
+		return "", fmt.Errorf("property %q has type json, which is not supported in data warehouse exports", path)
 	}
 
 	// Validate the operator for the property kind.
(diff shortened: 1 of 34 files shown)


core/filters.go
```diff
@@ -424,263 +491,387 @@ func retrieveFilterProperty(properties types.Properties, path string) (types.Pro
...
+				return nil, err
+			}
+			rules = append(rules, group)
+		case isCondition:
+			if err := validateFilterJSONProperties(rawRule, "property", "operator", "values"); err != nil {
+				return nil, err
+			}
+			_, hasConditionOperator := rawRule.Get([]string{"operator"})
+			values, hasValues := rawRule.Get([]string{"values"})
+			if !hasConditionOperator || !hasValues {
```
zapateo: In this branch, it is mandatory to pass _values_ ​​even for conditions that logically don't have any.

Do we want to restore the `main` behavior, or introduce this change?

Regardless of what we choose, the filter documentation in the API documentation on krenalis.com (for example [this](https://www.krenalis.com/docs/api/pipelines/import-users-from-applications#toc-h3-1)) needs to be updated accordingly.

core/filters.go
zapateo: Remember to update the API documentation on krenalis.com regarding the filter structure, which can now include group conditions.

core/filters.go
```diff
@@ -424,263 +491,387 @@ func retrieveFilterProperty(properties types.Properties, path string) (types.Pro
...
+	}
+	if !rulesValue.IsArray() {
+		return nil, errors.New("filter rules must be an array")
+	}
+
+	var operator FilterLogical
+	if err := operatorValue.Unmarshal(&operator); err != nil {
+		return nil, err
+	}
+	rules := make([]FilterRule, 0, rulesValue.NumElement())
```
zapateo: `rules` is allocated with the number of elements in `rulesValue` _before_ checking whether that number is within the allowed limits (the `*ruleCount > maxFilterRuleCount` check below).

Won't this cause allocation problems?

core/internal/filters/applies.go
```diff
@@ -18,78 +18,91 @@ func Applies(where *state.Where, attributes map[string]any) bool {
...
-			applies = exists && opIsNull(v)
-		case state.OpIsNotNull:
-			applies = !(exists && opIsNull(v))
-		case state.OpExists:
-			applies = exists
-		case state.OpDoesNotExist:
-			applies = !exists
+		switch rule := rule.(type) {
+		case *state.Where:
+			applies = Applies(rule, attributes)
```
zapateo: Here, the exported function `Applies` calls itself recursively. We generally tend to avoid doing this, introducing a not-exported function, or am I wrong?

However, if this solution is more practical, I'm fine with leaving it as is. It was just a matter of pointing this out; you decide.

core/filters_test.go
```diff
@@ -471,45 +659,7 @@ func Test_parseUint(t *testing.T) {
 	}
 }
 
-func Test_parseUUID(t *testing.T) {
```
zapateo: This test (`Test_parseUUID`) has been removed from this branch.

Have its tests been merged into some other test, or has it been removed entirely?

core/filters.go
```diff
@@ -424,263 +491,387 @@ func retrieveFilterProperty(properties types.Properties, path string) (types.Pro
...
+}
 
-	// disallowJSON indicates if "json" properties are disallowed.
-	disallowJSON := role == state.Destination && target == state.TargetUser
-
-	// disallowEmptyOnObject indicates if "is empty" and "is not empty" operators are disallowed on object properties.
-	disallowEmptyOnObject := role == state.Destination && target == state.TargetUser || target == state.TargetEvent
+// validateFilterCondition checks the validity of a filter condition and
+// returns its property path.
+func validateFilterCondition(cond *FilterCondition, validation *filterValidation) ([]string, error) {
```
zapateo: This function returns a `[]string`. Why not just `string`? Are there cases where the return value contains more than one element?

At one point, I wondered if the intended return was the individual path parts (e.g., `traits`, `email` instead of `traits.email`), but seeing how the returned value is used:

https://github.com/krenalis/krenalis/blob/f61c48be4066220b667bab3d0d153ae55eb80faf/core/filters.go#L851-L859

It seems like each path is a separate string, and therefore a simple `string` should be returned here.

Is there something I'm not understanding?

zapateo (changes requested):

zapateo (changes requested):
Please see the open conversations.

core/filters.go
gazerro: Fixed in commit [Return a single path from filter condition validation](https://github.com/krenalis/krenalis/pull/2427/changes/8d886eb0825768d333574aaf143a485208eda5cf).

zapateo (approved):

PR #2432, opened 2026-08-22
warehouses: preserve profiles after failed Identity Resolution

gazerro:
```
warehouses: preserve profiles after failed Identity Resolution (#2432)

Publish a candidate profiles version only after Identity Resolution
succeeds, preserve the visible profiles on failure, and remove obsolete
failed candidates without interfering with operations that are still
running.

Use the last published profiles version for schema changes and as the
source of new candidates after a failed run. Continue allocating from
the highest recorded version so failed or running candidates are not
reused.

Reconcile ambiguous transaction commit responses with the operation
state stored in the warehouse, so retries preserve the persisted success
or failure. Limit warehouse operation error messages before they reach
Krenalis resources or logs.

Cover these recovery and schema-version paths with aligned PostgreSQL
and Snowflake integration tests.
```

### DIFF
--- a/warehouses/postgresql/identityresolution.go
+++ b/warehouses/postgresql/identityresolution.go
@@ -46,9 +46,9 @@ func (warehouse *PostgreSQL) ResolveIdentities(ctx context.Context, opID string,
 		err = warehouse.finalizeIdentityResolution(ctx, pool, opID, err)
 	}()
 
-	// Determine the current version of the "krenalis_profiles" table and create a copy
-	// of it with the incremented version.
-	profilesVersion, err := warehouse.profilesVersion(ctx)
+	// Determine the greatest recorded version of the "krenalis_profiles" table
+	// and allocate the next version.
+	maxProfilesVersion, err := warehouse.maxProfilesVersion(ctx)
 	if err != nil {
 		return err
 	}
@@ -56,14 +56,14 @@ func (warehouse *PostgreSQL) ResolveIdentities(ctx context.Context, opID string,
 	if err != nil {
 		return err
 	}
-	newProfilesVersion := profilesVersion + 1
+	newProfilesVersion := maxProfilesVersion + 1
 	newProfilesName := fmt.Sprintf("krenalis_profiles_%d", newProfilesVersion)
 
 	// Prepare a candidate profiles version without exposing partial Identity
 	// Resolution results.
 	err = warehouse.execTransaction(ctx, func(tx pgx.Tx) error {
 		removePreviousProfilesTable := false
-		if profilesVersion != publishedProfilesVersion {
+		if maxProfilesVersion != publishedProfilesVersion {
 			// The latest unpublished profiles version may belong to an operation
 			// that is still running. Replace it only if that operation failed.
 			err = tx.QueryRow(ctx, `SELECT EXISTS (
@@ -73,17 +73,18 @@ func (warehouse *PostgreSQL) ResolveIdentities(ctx context.Context, opID string,
 				WHERE "v"."version" = $1
 					AND "o"."completed_at" IS NOT NULL
 					AND "o"."error" <> ''
-			)`, profilesVersion).Scan(&removePreviousProfilesTable)
+			)`, maxProfilesVersion).Scan(&removePreviousProfilesTable)
 			if err != nil {
 				return err
 			}
 			if !removePreviousProfilesTable {
 				return fmt.Errorf(
-					"profiles version %d is unpublished but does not belong to a failed operation", profilesVersion)
+					"profiles version %d is unpublished but does not belong to a failed operation", maxProfilesVersion)
 			}
 		}
-		// Create the candidate table by copying the schema from the current profiles table.
-		_, err := tx.Exec(ctx, fmt.Sprintf(`CREATE TABLE %s (LIKE "krenalis_profiles_%d")`, quoteIdent(newProfilesName), profilesVersion))
+		// Create the candidate table by copying the schema from the published profiles table.
+		_, err := tx.Exec(ctx, fmt.Sprintf(`CREATE TABLE %s (LIKE "krenalis_profiles_%d")`,
+			quoteIdent(newProfilesName), publishedProfilesVersion))
 		if err != nil {
 			return fmt.Errorf("cannot create profiles table (with name %s): %s", quoteIdent(newProfilesName), err)
 		}
@@ -95,8 +96,8 @@ func (warehouse *PostgreSQL) ResolveIdentities(ctx context.Context, opID string,
 			return err
 		}
 		if removePreviousProfilesTable {
-			// Drop the failed candidate table after preserving its schema changes.
-			_, err = tx.Exec(ctx, `DROP TABLE IF EXISTS "krenalis_profiles_`+strconv.Itoa(profilesVersion)+`"`)
+			// Drop the failed candidate table.
+			_, err = tx.Exec(ctx, `DROP TABLE IF EXISTS "krenalis_profiles_`+strconv.Itoa(maxProfilesVersion)+`"`)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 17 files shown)


warehouses/postgresql/identityresolution.go
```diff
@@ -24,59 +24,83 @@ import (
...
+	// Resolution results.
 	err = warehouse.execTransaction(ctx, func(tx pgx.Tx) error {
+		removePreviousProfilesTable := false
+		if profilesVersion != publishedProfilesVersion {
+			// The latest unpublished profiles version may belong to an operation
+			// that is still running. Replace it only if that operation failed.
+			err = tx.QueryRow(ctx, `SELECT EXISTS (
+				SELECT 1
+				FROM "krenalis_profile_schema_versions" "v"
+				JOIN "krenalis_system_operations" "o" ON "o"."id" = "v"."operation"
```
zapateo: The comment on this line should be updated:

https://github.com/krenalis/krenalis/blob/a9fe03f5517d4ee0f05aa211ca66bac55fa7bef7/warehouses/postgresql/tables/profile_schema_versions.sql#L7

as `operation` is no longer used just for logging purposes.

Same for Snowflake:

https://github.com/krenalis/krenalis/blob/2dffa1ee55516c34c2d123df8a63014352f30a39/warehouses/snowflake/tables/profile_schema_versions.sql#L7

warehouses/snowflake/identityresolution.go
```diff
@@ -205,20 +238,104 @@ func (warehouse *Snowflake) resolveIdentities(ctx context.Context, opID string,
 		return snowflake(err)
 	}
 
-	// Replace the current "profiles" view with a new one using the "CREATE OR
-	// REPLACE VIEW" statement since the table "_profiles" that the view refers to
-	// has changed its name.
-	_, err = db.ExecContext(ctx, createViewQuery(newProfilesName, profileColumns, true))
+	// Committing the successful operation switches the staged view to the new
+	// profiles without requiring another DDL statement.
+	err = warehouse.execTransaction(ctx, func(tx *sql.Tx) error {
```
zapateo: Suppose the IR has completed successfully, the UPDATE of `setOperationAsCompleted` runs successfully on Snowflake, but a network error occurs, and `setOperationAsCompleted` returns an error.

The code here returns, and this defer is then executed:

https://github.com/krenalis/krenalis/blob/3a4a4564c48dbaf054fccc1b8d1517a7d057f385/warehouses/snowflake/identityresolution.go#L45-L47

Krenalis notices that the operation was successful and restores the operation.

However, the fact remains that this code is not executed:

https://github.com/krenalis/krenalis/blob/3a4a4564c48dbaf054fccc1b8d1517a7d057f385/warehouses/snowflake/identityresolution.go#L250-L261

Is what I'm saying correct? So when will these cleanups be performed?

In PostgreSQL this does not appear to happen because these operations occur within a transaction.

warehouses/postgresql/operations.go
```diff
@@ -25,91 +26,98 @@ const (
...
-// nothing.
-func (warehouse *PostgreSQL) setOperationAsCompleted(ctx context.Context, opID string, opError error) error {
-	pool, _, err := warehouse.connectionPool(ctx, false)
+// readOperationStatus reads the persisted status of an operation. It returns
+// nil when the operation does not exist.
+func (warehouse *PostgreSQL) readOperationStatus(ctx context.Context, conn connection, opID string) (*opStatus, error) {
+
+	var completedAt *time.Time
+	var opError string
+	err := conn.QueryRow(ctx, `SELECT "completed_at", LEFT("error", $2)`+
```
zapateo: The [LEFT](https://www.postgresql.org/docs/9.1/functions-string.html) function in PostgreSQL operates on characters, not bytes:

> Return the first n characters in the string.

Similarly, in Snowflake, the [LEFT](https://docs.snowflake.com/en/sql-reference/functions/left) function also operates on characters when the argument (as in this case) is a VARCHAR:

> The number of UTF-8 characters to return if the input is a VARCHAR value.

However, the following parameter is passed:

```
warehouses.MaxOperationErrorBytes
```

which expresses the length in bytes.

I don't know if this causes problems later, because in any case the length is checked again in `warehouses/warehouses.go`, but it seems like an inaccuracy that could cause confusion. What do you think?

warehouses/postgresql/warehouse.go
```diff
@@ -439,6 +439,27 @@ func (warehouse *PostgreSQL) profilesVersion(ctx context.Context) (int, error) {
...
+	err = pool.QueryRow(ctx, `SELECT COALESCE(MAX("v"."version"), 0)
+		FROM "krenalis_profile_schema_versions" "v"
+		JOIN "krenalis_system_operations" "o" ON "o"."id" = "v"."operation"
+		WHERE "o"."completed_at" IS NOT NULL AND "o"."error" = ''`).Scan(&version)
+	if err != nil {
+		return 0, err
+	}
+	if version < 0 {
+		return 0, fmt.Errorf("warehouse returned a negative published profile schema version")
+	}
```
zapateo: Is this check really necessary? If it were not, I fear it could introduce confusion, suggesting the possibility of some cases that in reality do not occur.

Because the query returns the value of `COALESCE(MAX("v"."version"), 0)`, which I don't think can ever be less than 0, and furthermore, we don't perform this check in the method just above (`profilesVersion`):

https://github.com/krenalis/krenalis/blob/3a4a4564c48dbaf054fccc1b8d1517a7d057f385/warehouses/postgresql/warehouse.go#L434-L439

The same applies to Snowflake.

zapateo (changes requested):

warehouses/postgresql/identityresolution.go
gazerro: Fixed in commit [Clarify profile schema version operation comments](https://github.com/krenalis/krenalis/pull/2432/changes/2f33c1fab21d890ea0ef79ad2263ff3be071bd11).

warehouses/snowflake/identityresolution.go
gazerro: Fixed in commit [Continue Identity Resolution cleanup after ambiguous commits](https://github.com/krenalis/krenalis/pull/2432/changes/1ad5a716e0750f8ae24ae405ecb4b4b85c4f3cfa).

warehouses/postgresql/operations.go
gazerro: Good catch. This does not allow an oversized error through because `NewPersistedOperationError` still enforces the exact byte limit in Go, while `LEFT` limits the query result as an extra precaution. I made the units explicit with `operationErrorReadLimitCharacters` and documented the distinction.

See commit [Clarify operation error read limit units](https://github.com/krenalis/krenalis/pull/2432/changes/232f53703ce998d6233dc1f713193a287e6cc9d0).

warehouses/postgresql/warehouse.go
gazerro: Good point about the inconsistency. `COALESCE` only returns `0` when `MAX` is `NULL`; it can still return a negative value if the warehouse metadata contains one. Since values read from the warehouse are untrusted, I kept the check and added the same validation to `maxProfilesVersion` in both PostgreSQL and Snowflake. 

The method comments now document the non-negative return contract.

Fixed in commit [Validate maximum profile schema versions](https://github.com/krenalis/krenalis/pull/2432/changes/bbb7fcff14bf6bc9a6bb0fc6d4940843eb4dc4f6).

warehouses/snowflake/identityresolution.go
zapateo: On Snowflake, after a successful Identity Resolution, the DROP of the previous profile table is a best-effort operation: if it fails, the error is simply logged, and in some cases the cleanup isn't performed at all.

That table isn't deleted anywhere else, so it stays there forever: it's a complete copy of the profiles, so it takes up significant space, and each new occurrence leaves another.

zapateo (changes requested):

warehouses/snowflake/identityresolution.go
gazerro: Fixed in commit [Make Snowflake Identity Resolution cleanup safe to retry](https://github.com/krenalis/krenalis/pull/2432/changes/ff8ff85ef6f836b80e7b7ef1ba6a4505aa3a51c3).

zapateo (approved):

PR #2439, opened 2026-08-23
core: fix batching across application record pages

gazerro:
```
core: fix batching across application record pages (#2439)

Prevent page boundaries from marking records on subsequent pages as
final, which causes imports and exports to flush one record at a time
after the first page.

Simplify record iteration by removing final-record tracking and
completing pending batches after iteration ends.
 ```


zapateo (changes requested):
Please, resolve the merge conflicts.

zapateo (approved):

PR #2446, opened 2026-08-27
tools/backoff: add retry state reset

gazerro:
```
tools/backoff: add retry state reset (#2446)

Add Reset to restart a backoff sequence while preserving its
configuration. Reset clears the attempt count and next wait duration.

Reset must not be called while an AfterFunc callback is pending.
```

### DIFF
--- a/tools/backoff/backoff.go
+++ b/tools/backoff/backoff.go
@@ -134,6 +134,14 @@ func (bo *Backoff) Next(ctx context.Context) bool {
 	return true
 }
 
+// Reset clears the attempt counter and next wait time while preserving the
+// configured attempts, base, and cap. It must not be called while an AfterFunc
+// callback is pending.
+func (bo *Backoff) Reset() {
+	bo.attempt = 0
+	bo.waitTime = 0
+}
+
 // SetAttempts sets the attempts. Use backoff.NoLimit for unlimited attempts.
 // It panics if attempts is zero or negative.
 func (bo *Backoff) SetAttempts(attempts int) {
--- a/tools/backoff/backoff_test.go
+++ b/tools/backoff/backoff_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"math"
 	"testing"
+	"testing/synctest"
 	"time"
 )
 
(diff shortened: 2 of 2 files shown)


tools/backoff/backoff.go
zapateo: This isn't related to this PR itself; I'm writing it to you so we can fix it on `main`.

`NoLimit` is referenced in three places (two in the tests and one in the documentation), but this constant doesn't exist anywhere.

Feel free to close this conversation once you read it.

zapateo: If I understand backoff correctly, a test like this should succeed (because I expect `Reset` to reset backoff to an initial state), but it fails.

```go
func Test_Reset_Timer(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	bo := New(1)

	// Arm the timer with a wait long enough that only the cancellation can end
	// the second Next.
	bo.SetNextWaitTime(time.Hour)
	cancel()
	if bo.Next(ctx) {
		t.Fatal("Next returned true on a canceled context")
	}

	bo.Reset()

	// AfterFunc was never called, so Stop must report that no function is
	// scheduled, exactly as it does on a new Backoff.
	if got, want := bo.Stop(), New(1).Stop(); got != want {
		t.Fatalf("Stop after Reset: got %v, want %v (Reset left the timer armed)", got, want)
	}
}
```

What do you think? Am I misunderstanding something?

zapateo: It's possible that with this change, some points in the code (indicated below) could take advantage of the `Reset` method.

Since they're not specifically within the scope of this PR, they're fine to leave as they are (provided that the `Reset` method can be actually used).

You decide what to do.

https://github.com/krenalis/krenalis/blob/d07477292933b33830479159fd80eb4f3a0aecbd/core/internal/datastore/flusher.go#L183

https://github.com/krenalis/krenalis/blob/40729fc7d14a2de558cbc2d520262e8543a0a19b/core/core.go#L1514

zapateo (changes requested):

tools/backoff/backoff.go
gazerro: Yes, that's exactly the purpose of this pull request. I was thinking of opening another PR once this one has been merged.

gazerro: Even if you comment out the line

`bo.Reset()`

`bo.Stop()` still returns `true` instead of `false`.

This happens because `bo.Next(ctx)` doesn’t reset the timer or set it to `nil` when the context has expired.

So this should probably be investigated and, if necessary, fixed in a separate PR.

gazerro: I'll investigate this and fix it in a separate PR.

zapateo (approved):

PR #2447, opened 2026-08-27
core/internal/datastore: fix object property name reuse

gazerro:
```
core/internal/datastore: fix object property name reuse (#2447)

When a deleted object property is replaced by a property with the same
name, drop the deleted property's flattened warehouse columns before
adding the replacement.

This prevents schema alterations from trying to drop the deleted
property's logical name, which is not a warehouse column.
```

### DIFF
--- a/core/internal/datastore/diffschemas/diffschemas.go
+++ b/core/internal/datastore/diffschemas/diffschemas.go
@@ -230,6 +230,7 @@ func Diff(oldSchema, newSchema types.Type, rePaths map[string]any, path string) 
 		}
 
 		if v, ok := rePaths[keptPath]; ok && v == nil {
+			columnBase := pathToColumn(keptPath)
 			if renamed {
 				// New properties with the same name of a renamed property. They
 				// appear in "rePaths" (the key is the name of the created
@@ -249,13 +250,13 @@ func Diff(oldSchema, newSchema types.Type, rePaths map[string]any, path string) 
 					for _, p := range propertyPaths(oldProp.Type) {
 						operations = append(operations, warehouses.AlterOperation{
 							Operation: warehouses.OperationDropColumn,
-							Column:    pathToColumn(appendPath(path, appendPath(keptName, p))),
+							Column:    columnBase + "_" + pathToColumn(p),
 						})
 					}
 				} else {
 					operations = append(operations, warehouses.AlterOperation{
 						Operation: warehouses.OperationDropColumn,
-						Column:    pathToColumn(keptPath),
+						Column:    columnBase,
 					})
 				}
 			}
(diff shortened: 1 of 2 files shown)


core/internal/datastore/diffschemas/diffschemas.go
zapateo: The current fix fails with such a case (add this in `core/internal/datastore/diffschemas/diffschemas_test.go`):

```go
{
	name: "One nested object property removed and then an object property added with the same name",
	fromSchema: types.Object([]types.Property{
		{Name: "x", Type: types.Object([]types.Property{
			{Name: "a", Type: types.Object([]types.Property{
				{Name: "b", Type: types.String(), Nullable: true},
				{Name: "c", Type: types.Int(32), Nullable: true},
			})},
		})},
	}),
	toSchema: types.Object([]types.Property{
		{Name: "x", Type: types.Object([]types.Property{
			{Name: "a", Type: types.Object([]types.Property{
				{Name: "d", Type: types.String(), Nullable: true},
			})},
		})},
	}),
	rePaths: map[string]any{"x.a": nil},
	expectedOps: []warehouses.AlterOperation{
		{Operation: warehouses.OperationDropColumn, Column: "x_a_b"},
		{Operation: warehouses.OperationDropColumn, Column: "x_a_c"},
		{Operation: warehouses.OperationAddColumn, Column: "x_a_d", Type: types.String()},
	},
},
```

The problem seems to be that `appendPath` shouldn't be used, because `keptPath` already contains the path. If you compare the expected output with the output from the test above, you'll see the problem. The fix should be simply:

```diff
diff --git a/core/internal/datastore/diffschemas/diffschemas.go b/core/internal/datastore/diffschemas/diffschemas.go
index 3163fbe9d..618810ad8 100644
--- a/core/internal/datastore/diffschemas/diffschemas.go
+++ b/core/internal/datastore/diffschemas/diffschemas.go
@@ -263,7 +263,7 @@ func Diff(oldSchema, newSchema types.Type, rePaths map[string]any, path string)
 				for _, c := range util.PropertiesToColumns(newProp.Type.Properties()) {
 					operations = append(operations, warehouses.AlterOperation{
 						Operation: warehouses.OperationAddColumn,
-						Column:    pathToColumn(appendPath(path, keptPath)) + "_" + c.Name,
+						Column:    pathToColumn(keptPath) + "_" + c.Name,
 						Type:      c.Type,
 					})
 				}

```

zapateo (changes requested):

core/internal/datastore/diffschemas/diffschemas.go
gazerro: Fixed in commit [Fix nested object recreation](https://github.com/krenalis/krenalis/pull/2447/changes/650b8b82e896c53bdd452a161bcfdac895c61e3b).

See also commit [Reduce allocations when recreating properties](https://github.com/krenalis/krenalis/pull/2447/changes/098418e28e2bc46c76a5f2b044897bac594b78c7).

zapateo (approved):

PR #2448, opened 2026-08-27
Property display name

gazerro:


tools/types/types.go
zapateo: The default profile schema, the one used when creating a workspace by the Admin, uses description fields to store what are essentially the display names of the properties.

The schema is found in the file: `admin/src/components/routes/WorkspaceCreate/InitialSchema.json`

I think it should be updated to use display names instead of descriptions.

zapateo: Consider this situation:

- `first_name` has a display name but no description
- `last_name` has a description but no display name

In the mappings, the two properties are currently displayed like this:

Image

When I expected both the display name and the description to be displayed (perhaps next to them, or something like that).

The same problem currently occurs in full mode as well:

Image

zapateo: Currently, this change doesn't affect any connectors, but some could be modified to use the display name instead of the description.

Just to give one example among many:

https://github.com/krenalis/krenalis/blob/41130291497562883954b7ea2272aace4a085bb7/connectors/googleanalytics/event_types.go#L43-L46

This raises two questions:

- Should we also modify the connectors? In this PR or another?
- Should we modify the connector generation SKILLS, as well as krenalis.com, to indicate the use of displayed names?

zapateo: The documentation on krenalis.com needs to be updated:

At least these two pages; I don't know if there are others:

- https://www.krenalis.com/docs/api/profile-schema
- https://www.krenalis.com/docs/api/schemas

zapateo (changes requested):

tools/types/types.go
gazerro: Yes, I was considering that as a separate change, since the default schema will also need to be updated for roles, and the transformation screens will need to be updated to show the display name as well.

Given that these changes touch multiple areas, I think it would be better to handle them in separate PRs rather than include them in this one. This PR should focus only on adding the display name to properties and updating the profile schema editing flow in the Admin.

gazerro: See my previous comment. I'll keep your feedback in mind for future PRs.

gazerro: See my previous comment. I'll keep your feedback in mind for future PRs.

gazerro:
See also commit [Test property display names](https://github.com/krenalis/krenalis/pull/2448/changes/2e0c9ccd650740d2dc5744d6620ea877112b1061).

tools/types/types.go
zapateo: As discussed verbally, we should:

*   Update the branch base for the documentation so that it targets the next release (`v0.44.0`) rather than `main`.

*   Update the Krenalis version within the branch to point to this specific branch, ensuring the profile schema documentation is updated as well. The procedure is documented [here](https://github.com/krenalis/krenalis.com#need-a-newer-krenalis-version), though it needs to be slightly modified to point to this branch.

zapateo (changes requested):

tools/types/types.go
gazerro: Done: https://github.com/krenalis/krenalis.com/pull/56.

zapateo (approved):

PR #2452, opened 2026-08-31
core/internal/datastore/diffschemas: fix drop of a reused object property

zapateo:
## Commit message

```
core/internal/datastore/diffschemas: fix drop of reused object property (#2452)

This commit is similar to e4fd56faf (#2447, "fix object property name
reuse"), with the difference that #2447 handled the drop of a deleted
object property whose name is reused by a newly created property, while
this commit handles the same drop when the name is instead reused by a
renamed property.

When a property is deleted and the name it frees is taken over by a
renamed property, Diff emits a single DropColumn for the freed path. If
the deleted property is an object, no column with that name exists in
the warehouse, since object properties are flattened into one column per
leaf. The real columns are therefore never dropped, and the drop targets
a column that does not exist. The RenameColumn operations emitted right
after take over those same leaf columns, which are still occupied.

Drop one column per leaf of the deleted property, the same handling
#2447 added for a name reused by a new property rather than by a rename.
```

### DIFF
--- a/core/internal/datastore/diffschemas/diffschemas_test.go
+++ b/core/internal/datastore/diffschemas/diffschemas_test.go
@@ -708,6 +708,68 @@ func TestDiff(t *testing.T) {
 			},
 		},
 		{
+			name: "Object property deleted and its name reused by a renamed object property",
+			fromSchema: types.Object([]types.Property{
+				{Name: "foo", Type: types.Object([]types.Property{
+					{Name: "b", Type: types.String(), Nullable: true},
+					{Name: "c", Type: types.String(), Nullable: true},
+				})},
+				{Name: "bar", Type: types.Object([]types.Property{
+					{Name: "b", Type: types.String(), Nullable: true},
+					{Name: "c", Type: types.String(), Nullable: true},
+				})},
+			}),
+			toSchema: types.Object([]types.Property{
+				{Name: "foo", Type: types.Object([]types.Property{
+					{Name: "b", Type: types.String(), Nullable: true},
+					{Name: "c", Type: types.String(), Nullable: true},
+				})},
+			}),
+			rePaths: map[string]any{"foo": "bar"},
+			expectedOps: []warehouses.AlterOperation{
+				{Operation: warehouses.OperationDropColumn, Column: "foo_b"},
+				{Operation: warehouses.OperationDropColumn, Column: "foo_c"},
+				{Operation: warehouses.OperationRenameColumn, Column: "bar_b", NewColumn: "foo_b"},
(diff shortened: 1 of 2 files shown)


core/internal/datastore/diffschemas/diffschemas_test.go
```diff
@@ -771,6 +771,68 @@ func TestDiff(t *testing.T) {
 				{Operation: warehouses.OperationAddColumn, Column: "x_a", Type: types.String()},
 			},
 		},
+		{
```
gazerro: In the tests, the deleted object and the renamed object have exactly the same leaves. As a result, the tests do not verify that the list of columns to delete is actually derived from `oldProp.Type`: an incorrect implementation based on `newProp.Type` would still pass.

For example, you could add the following test:

```go
{
			name: "Object property deleted and its name reused by a differently shaped renamed object property",
			fromSchema: types.Object([]types.Property{
				{Name: "foo", Type: types.Object([]types.Property{
					{Name: "a", Type: types.String(), Nullable: true},
					{Name: "b", Type: types.String(), Nullable: true},
				})},
				{Name: "bar", Type: types.Object([]types.Property{
					{Name: "c", Type: types.String(), Nullable: true},
					{Name: "d", Type: types.Object([]types.Property{
						{Name: "e", Type: types.String(), Nullable: true},
					})},
				})},
			}),
			toSchema: types.Object([]types.Property{
				{Name: "foo", Type: types.Object([]types.Property{
					{Name: "c", Type: types.String(), Nullable: true},
					{Name: "d", Type: types.Object([]types.Property{
						{Name: "e", Type: types.String(), Nullable: true},
					})},
				})},
			}),
			rePaths: map[string]any{"foo": "bar"},
			expectedOps: []warehouses.AlterOperation{
				{Operation: warehouses.OperationDropColumn, Column: "foo_a"},
				{Operation: warehouses.OperationDropColumn, Column: "foo_b"},
				{Operation: warehouses.OperationRenameColumn, Column: "bar_c", NewColumn: "foo_c"},
				{Operation: warehouses.OperationRenameColumn, Column: "bar_d_e", NewColumn: "foo_d_e"},
			},
		},
```

This requires `fix-type-checks-on-renames` to be merged first, so I suggest adding the test afterward.

gazerro (changes requested):

core/internal/datastore/diffschemas/diffschemas_test.go
zapateo: As discussed verbally, let's close this PR first and then add this test later, along with the other fix; it seems like the easiest and quickest way to proceed.

gazerro (approved):

PR #2453, opened 2026-08-31
core/internal/datastore/diffschemas: fix type check on renames

zapateo:
## Commit message

```
core/internal/datastore/diffschemas: fix type check on renames (#2453)

When a property is deleted and the name it frees is taken over by a
renamed property, Diff compares the type of the deleted property with
the new one, instead of the type of the property being renamed. A type
change on the renamed property therefore goes unnoticed, and the
RenameColumn operation leaves in the warehouse a column with the old
type under a property declared with the new one. Conversely, a rename
onto the name of a deleted property of a different type is rejected
with a misleading error.

Compare the property at the old path with the new one, as already done
for renames to a name that did not exist in the old schema. Report the
error on the renamed property, naming both its old and its new path.
```

### DIFF
--- a/core/internal/datastore/diffschemas/diffschemas.go
+++ b/core/internal/datastore/diffschemas/diffschemas.go
@@ -278,8 +278,13 @@ func Diff(oldSchema, newSchema types.Type, rePaths map[string]any, path string) 
 				Operation: warehouses.OperationDropColumn,
 				Column:    pathToColumn(keptPath),
 			})
-			if !types.Equal(oldProp.Type, newProp.Type) {
-				return nil, fmt.Errorf("error on property %q: type changes are not supported", appendPath(path, oldProp.Name))
+			// The type to compare with the new one is that of the renamed
+			// property, at oldPath, and not that of the deleted property, whose
+			// name is the one that has been reused.
+			oldName := propPathToName(oldPath)
+			renamedProp, _ := oldProperties.ByName(oldName)
+			if !types.Equal(renamedProp.Type, newProp.Type) {
+				return nil, fmt.Errorf("error on property %q (renamed to %q): type changes are not supported", appendPath(path, oldName), keptPath)
 			}
 			if newProp.Type.Kind() == types.ObjectKind {
 				for _, c := range util.PropertiesToColumns(newProp.Type.Properties()) {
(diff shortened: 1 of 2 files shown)


core/internal/datastore/diffschemas/diffschemas.go
```diff
@@ -296,8 +296,13 @@ func Diff(oldSchema, newSchema types.Type, rePaths map[string]any, path string)
 					Column:    pathToColumn(keptPath),
 				})
 			}
-			if !types.Equal(oldProp.Type, newProp.Type) {
-				return nil, fmt.Errorf("error on property %q: type changes are not supported", appendPath(path, oldProp.Name))
+			// The type to compare with the new one is that of the renamed
```
gazerro:
```
  // Compare the new type with the type of the property renamed from oldPath,
  // rather than with the deleted property whose name was reused.
```

core/internal/datastore/diffschemas/diffschemas_test.go
```diff
@@ -771,6 +771,79 @@ func TestDiff(t *testing.T) {
...
+				{Name: "foo", Type: types.String(), Nullable: true},
+			}),
+			rePaths: map[string]any{"foo": "bar"},
+			expectedOps: []warehouses.AlterOperation{
+				{Operation: warehouses.OperationDropColumn, Column: "foo"},
+				{Operation: warehouses.OperationRenameColumn, Column: "bar", NewColumn: "foo"},
+			},
+		},
+		{
+			name: "Object property deleted and its name reused by a differently shaped renamed object property",
```
gazerro: The current test name is slightly ambiguous because it may suggest that the renamed object changed shape during the rename. Consider making it explicit that its shape differs from that of the deleted object:

> Deleted object property name reused by a renamed object property whose shape differs from the deleted one

gazerro (changes requested):

core/internal/datastore/diffschemas/diffschemas.go
zapateo: Done in [core/internal/datastore/diffschemas: clarify rename type-check comment](https://github.com/krenalis/krenalis/pull/2453/commits/d6f4111765e4574fca2d4b64eb583cbcdbff3552).

core/internal/datastore/diffschemas/diffschemas_test.go
zapateo: Done in [core/internal/datastore/diffschemas: clarify reused-name test case name](https://github.com/krenalis/krenalis/pull/2453/commits/87dee8b511e9bd6abb5fdb49651be74e70860d30).

gazerro (approved):

